PR #123, opened 2014-03-04
Remove ignore errors option on xs-authserver install and specify version to install

migonzalvar:
Two changes:

- remove ignore errors option on xs-authserver because the issue when re running the task is solved upstream
- specify version number to install


davelab6:
This seems very old. Should it be closed?

PR #153, opened 2014-07-05
Feature dnsmasq

scollazo:
This PR includes the following changes:
- Replaces bind with dnsmasq
- Creates a configuration option for opendns (this feature was missing from xsce)
- Creates a configuration option for google safe search enforcement.

Dnsmasq works as a dns forwarder and cache. This means that will query the dns servers specified in /etc/resolv.conf and cache the results,  instead of using the dns root servers as bind does.

This PR needs testing , I only tested it on a vm, and not with the real environmnet.


davelab6:
This seems very old. Should it be closed?

scollazo:
I don't think this is needed, closing.

holta:
fyi dnsmasq was just recently implemented as part of the dominant DNS-and-other tool for XSCE &mdash; since renamed to Internet-in-a-Box (IIAB):

e.g. in places like:
https://github.com/iiab/iiab/blob/master/roles/network/tasks/dnsmasq.yml
https://github.com/iiab/iiab/blob/master/roles/network/templates/network/dnsmasq-iiab
https://github.com/iiab/iiab/blob/master/roles/network/templates/network/dnsmasq.conf.j2
https://github.com/iiab/iiab/blob/master/roles/network/templates/network/dnsmasq.service.rh

Coincidentally work is ongoing here just this week, as a few nettlesome conflicts remain Ubuntu 18.04 to get dnsmasq started reliably at boot time:
https://github.com/iiab/iiab/issues/1320

PR #195, opened 2015-01-27
Feature KA Lite

tim-moody:


tim-moody:
xo4 test failed due to ansible version, retesting one line change

tim-moody:
now installs on both 64bit vm and xo4

PR #198, opened 2015-01-31
php xo fix

jvonau:


georgejhunt:
That would apparently fix TK's problem. rwtab is only a vague memory from a couple of years ago.  I'm glad you're still with us occationally

tim-moody:
yep, I should have asked sooner.
the alternative is to change 'empty   /var/lib/php' to 'empty   /var/lib/php/session' so session files don't accumulate.  what do you think?

jvonau:
your choice, but I wouldn't, be nice to have around for inspection should
be box ever get cracked.

PR #199, opened 2015-02-01
use detected ansible_distribution_version

jvonau:
ansible_distribution_release can vary, a stock F18 install has "Spherical Cow" the above is specific to XOs only

### DIFF
--- a/roles/1-prep/tasks/computed_vars.yml
+++ b/roles/1-prep/tasks/computed_vars.yml
@@ -86,5 +86,5 @@
 - name: Set mysqld service name to mysqld for fedora 18
   set_fact:
     mysql_service: mysqld
-  when: ansible_distribution_version == "18"
+  when: 'ansible_distribution_version == "18" or ansible_distribution_release == "based on Fedora 18"'
 
--- a/roles/1-prep/tasks/computed_vars.yml
+++ b/roles/1-prep/tasks/computed_vars.yml
@@ -85,6 +85,6 @@
     
 - name: Set mysqld service name to mysqld for fedora 18
   set_fact:
-    mysql_service: mysqld    
-  when: ansible_distribution_release == "based on Fedora 18"
-                  
+    mysql_service: mysqld
+  when: ansible_distribution_version == "18"
+


tim-moody:
don't know if this works. on f18 we have ansible 1.4 and this fact is supposed to be new in 1.6
however, if I look at the facts I see this variable does exist, but on my xo4 its value is 13.

jvonau:
weird, on the F18 non xo here:
rpm -q ansible
ansible-1.4.1-1.fc18.noarch
when I do:  ansible -m setup -i ansible_hosts localhost --connection=local

        "ansible_distribution": "Fedora",
        "ansible_distribution_release": "Spherical Cow",
        "ansible_distribution_version": "18",

strange... very strange.. bug in ansible or the XO? I'll see where that is
coming from...

jvonau:
ok here is what is going on,  ansible_distribution* is derived from the contents of /etc/redhat-release which is a symlink to /etc/fedora-release. On a stock fedora install the string should be: Fedora release <version> <nickname> , however olpc-os-builder plays around with that file while building the image using their release number instead. That is where the '13' is originating from for the XOs.

tim-moody:
you could put something in 1-prep/templates/local_facts.  I have always thought we should have a specific flag identifying an xo and we could pick up a value for host_os or something, which ansible doesn’t seem to have.

jvonau:
we already have the XO model identified via xsce_facts, is that no longer
valid?

tim-moody:
if you look at 1-prep tasks main it uses the same value you were trying to change to determine we are on an xo.  it never felt right to me as I’m guessing it doesn’t feel right to you.

# Put conditional includes for platforms here
- include: xo.yml
  when: ansible_distribution_release == "based on Fedora 18"

xsce_facts is not really used as it was replaced by local_facts

vnstat has the only use

- name: get facts if not present
  xsce_facts:
  when: xsce_wan_iface is not defined

but this will never be true as xsce_wan_iface was defined in default_vars

but you are right; the xo version logic should probably be in local_facts. (but not the rest of xsce_facts as they are now derived from ansible facts) and we should use it.

jvonau:
How about pushing this one, as it is dealing with a platform issue around
OS detection. Once the XO version is available, we can change the
conditional for this, xo.yml, vnstat, and get rid of xsce_facts. Network
detection is not pure ansible, the shell code move from xsce_facts to
network, but I still say that is cleaner than messing with ansible's
detection.

tim-moody:
The problem is that as it stands it will break xo installs, which was the reason to include it in the first place.  There aren’t any other f 18 installs going on so nothing is broken at the moment.

jvonau:
Please have a look at the second commit, that covers both situations. I
guess my F18 non-XO installs don't count for anything then and I'm wasting
my time.

tim-moody:
didn’t know you had any f 18 installs, but I’m not aware of anything that would break.

jvonau:
Without having the correct DB name for the OS how does one add the latest
and greatest features to an older install? Or are only fresh installs going
to benefit from the newest features?

tim-moody:
superceded

PR #201, opened 2015-02-01
[bigfix] don't depend on hardcoded path for git repo

jvonau:
git should be able to live anywhere, use the present working directory for the git location.

### DIFF
--- a/roles/1-prep/templates/local_facts.fact.j2
+++ b/roles/1-prep/templates/local_facts.fact.j2
@@ -1,6 +1,7 @@
 #!/bin/sh
 . /etc/xsce/xsce.env
-cd $XSCE_DIR
+GIT_DIR=`pwd`
+cd $GIT_DIR
 
 # while we're here untrack local vars
 git update-index --assume-unchanged vars/local_vars.yml
@@ -8,6 +9,8 @@ git update-index --assume-unchanged vars/local_vars.yml
 # get current version
 BRANCH=`git rev-parse --abbrev-ref HEAD`
 COMMIT=`git rev-parse --verify HEAD`
+
+cd $XSCE_DIR
 
 if [ -d /usr/lib64/php ]
 then


tim-moody:
This is a conversation George needs to be in.  I agree with the sentiment, but we also have to handle the case where a script is running outside of the git repo directory and needs to be able to find it.  That is the purpose of /etc/xsce/xsce.env.  We currently write that from default vars.  We could equally have discovered the location on the first run, but things get very messy if it moves around.  Offline install expects it to be in a known location.  The location of the git repo is arbitrary, but the documentation tells the installer where to put it.

jvonau:
So add George in, I can see the benefit of having the git repo on a usbkey
for offline installs. I can't seem to find the offline code, can you point
it out to me?

georgejhunt:
I have not put any of the offline stuff into XSCE.  Currently it's in https://github.com/georgejhunt/xsce --branch offline.  At an earlier time, my strategy was to capture all the online sources, and stick them somewhere line /opt/schoolserver/* so I could use them offline. At this point, my tinycore approach seems to be able to capture the results of "runansible" successfully. 
Putting stuff in a known place is good, because it lets us drill down when changes stimulated by Tim's console, are not working offline. But I've encountered failures of the colsole code when the git repo was not running from /opt/schoolserver/xsce, and I don't really object to this requirement.

jvonau:
The git repo locations already differ, when online George is using /opt/xsce
in xs-config while the current instructions are to use /opt/schoolserver/xsce
for the location of the git repo. Well I'll leave it in your hands.

tim-moody:
I admit I wasn’t totally comfortable with fixing the location of the repo, but it is easy to get into a chicken/egg situation.  And it may limit backwards compatibility, though you could just reclone to the new default and everything else would stay the same.  Also, the location is fixed, but it is soft-coded where I thought George wanted it.

Nevertheless, we can allow the git repo to be anywhere if we agree on the following:

1) when you run runansible, runtags, or install-console they must be in your cwd
2) if you clone the repo and run any of the above and then want to move the git repo, you must remove /etc/xsce

Then I think the following changes could be made:

runansible, runtags, and install-console are modified to assign cwd to xsce_dir and pass into ansible (the old version used to do something similar to this)
xsce_dir is removed from default_vars.yml
xsce_dir is added to config_vars.yml when it is first created
everything else continues to use xsce_dir to determine the location of the repo

I would be tempted to output a debug message in prep if xsce_dir is undefined, but the ansible run will probably just fail in that case anyway.

Comments?

tim-moody:
This is superseded by PR 208

PR #202, opened 2015-02-01
XO detection

jvonau:
as discussed in #199


tim-moody:
I think this works.  As I understand it, it uses "based on Fedora 18" to determine the correct version of mysql and xo_model to determine whether to do xo.yml.

jvonau:
This commit has nothing to do with mysql, it is the result of our conversation. I'm just dealing with the XO detection for xo.yml

PR #205, opened 2015-02-03
Repos2

jvonau:
I mistakenly assumed that all the latest rpms would live in xs-extras, that proved to be untrue.


tim-moody:
ready to merge?

jvonau:
I believe so, I get the latest rpms with this

PR #207, opened 2015-02-04
squid no longer has verb for log_access

georgejhunt:
Required to runansible to completion. Testing squid function still remains not done.  One step at a time


tim-moody:
is this f21 only or backwards compatible?

georgejhunt:
The default value for log_access was to allow before the directive was
removed, so I believe that this will be backwards compatible:

http://www.squid-cache.org/Doc/config/log_access/

PR #208, opened 2015-02-04
REVIEW - run git anywhere (but don't move it afterwards)

tim-moody:
Please review.  If this looks right will add to runtags and install-console and
remove xsce_dir from default_vars
and add to config_vars

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,6 +48,15 @@ Or you can use the previous manual process by typing
 
 * './runansible'
 
+Please note that it is possible to clone the git repo into a different directory than the one
+recommended above.  If you do so, it will be necessary to edit vars/local_vars.yml and add the line
+
+xsce_dir: <full path to your git repo clone>
+
+prior to running ansible.  
+
+Relocating the git repo after running ansible will cause problems.
+
 If you want to explore and get dirty with the code, please read the ``HACKING``
 file. You would probably want to go through the `ansible documentation`_ before diving into the
 playbooks. Documentation for creating plugins for the server is under
--- a/install-console
+++ b/install-console
@@ -1,14 +1,38 @@
 #!/bin/bash
 
+# if not the first run repo, location is here
+
+if [ -f /etc/xsce/xsce.env ]
+then
+  . /etc/xsce/xsce.env
+  cd $XSCE_DIR
+fi
+
+if [ ! -f xsce.yml ]
+then
+ echo "XSCE Playbook not found."
+ echo "Please run this command from the top level of the git repo."
+ echo "Exiting."
+ exit
+fi
+
 PLAYBOOK="xsce-base.yml"
 INVENTORY="ansible_hosts"
+SELINUX_BEFORE=""
+SELINUX_AFTER=""
 
-SELINUX_BEFORE=`cat /etc/selinux/config | gawk -F= '/^SELINUX=/{ print $2 }'`
+if [ -f /etc/selinux/config]
+then
+ SELINUX_BEFORE=`cat /etc/selinux/config | gawk -F= '/^SELINUX=/{ print $2 }'`
+fi
 
 export ANSIBLE_LOG_PATH="xsce-install.log"
 ansible-playbook -i $INVENTORY $PLAYBOOK --connection=local 
 
-SELINUX_AFTER=`cat /etc/selinux/config | gawk -F= '/^SELINUX=/{ print $2 }'`
+if [ -f /etc/selinux/config]
+then
+ SELINUX_AFTER=`cat /etc/selinux/config | gawk -F= '/^SELINUX=/{ print $2 }'`
+fi
 
 if [ "$SELINUX_BEFORE" != "$SELINUX_AFTER" ]; then
   echo "Rebooting ..."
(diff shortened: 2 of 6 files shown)


jvonau:
Which one takes control ? 
ARGS="--extra-vars '{\"xsce_dir\":\"$CWD\",\"xsce_domain\":\"$XSDOMAIN\"}'"  
or 
--extra-vars "xsce_dir=/opt/schoolserver/xsce xsce_domain=lan" 
at the bottom? Is that a fallback?

georgejhunt:
let me do a little research. I may need to do some changes to installer for
this to work with that code.  I'll do it tomorrow.

georgejhunt:
I looked at it -- simple changes, and only three -- go ahead

tim-moody:
oops, cut and paste error.  this strings at the bottom were test cases that should have been removed.

the real question is whether this the right algorithm.
it occurred to me over night that perhaps
if /etc/xsce/xsce.env exists and $XSCE_DIR/xsce.yml does not exist that that should just be an error and exit.

tim-moody:
I think that any time one of the install scripts runs successfully you should be able to use /etc/xsce/xsce.env to determine where it is.  

Unfortunately with this this approach you will no longer be able to use the variable xsce_dir in default_vars.yml.  

That is the big change being considered, is the git location defacto and to be discovered or is it predefined in a variable and you are required to install in the predefined location.  

Certainly the documentation will tell you where to install it, so we are only handling the case where someone ‘knows what they’re doing’ and puts it somewhere else.  Even in that case xsce_dir: <my git location> in local_vars.yml will fix the problem, so you can put git anywhere you like, but after the install if you move it you will get its location out of sync with what's in xsce.env and what’s in the consoles config_vars.yml file.

This seems to me a tradeoff between certainty and flexibility and I’m not sure what side I’m on.

tim-moody:
I think this is ready to go. It still requires xsce_dir to hold the correct value for the location of the git repo, but it adds some error checking.  It also includes the prep tag if the user didn't.

jvonau:
Let's think about this a second, that would require that the git repo is in
the same place when running changes from within the GUI-console. In that
situation the removable device's location may not be present resulting in
not finding the playbooks.

Think we can have it both ways, but this requires copying the git repo,
identify when ansible is launched outside of xsce_dir and populate xsce_dir
with the CWD.

How does this look? #209 Thoughts?

tim-moody:
George uses tar to populate the entire root file system, so it presupposes that an install, probably install-console has already been done before the directory was tarred up.  So for offline install all the inconsistencies were already resolved by the packager.  He also uses /opt/schoolserver/downloads, etc. as a home for all the dependent rpms, pip files, tar files, and anything else required to do a complete install.  I don’t think ansible is ever run from the removable media.

Correct me if I am wrong, George.

georgejhunt:
That's what is working now. But of course, if ansible is run again from the
console, offline, then hopefully the resources are still where they were
the first time when the image was created online.

georgejhunt:
No, not just up to console. My intention is to do the full runansible, and
have everything installed (but not necessarily enabled), so that the tarred
image has everything (and then the console can be used to set the enabled
flags, and rerun ansible)

The whole tar of both rootfs, and library is about 1.5GB on the stick.  The
whole install from raw hardware to functioning system takes about 2 minutes
on my slow NUC (2830)

tim-moody:
>whole install from raw hardware to functioning system takes about 2 minutes
>on my slow NUC (2830)

amazing!

tim-moody:
Tested successfully on 64bit.  On xo4, runs out of space, but not due to this PR.

tim-moody:
Just ran it successfully on single nic vm.

jvonau:
Yes I agree, a second or later run of ansible would use
/opt/schoolserver/{downloads,pip-packages} that has been pre-populated with
the first run of ansible online. Manually setting the use_cache and
no_network flags does the right thing when all the packages are present for
what you want to do. This is true even when run from a non stock location.
Works great, faster, can you help me further understand the use of these
flags? It's just a question of bootstrapping the install on a first run. We
can have /opt/schooserver/* populated before calling ansible or within a
playbook. Looks like shell scripting around this now, thought the push was
to use ansible.

jvonau:
when running from the GUI console:

********************
failed: [127.0.0.1] => {"failed": true, "item": ""}
msg: Configuration directory (/opt/schoolserver/xsce)does not match actual directory (/opt/schoolserver/xsce_cmdsrv). Please consult documentation. Exiting.

tim-moody:
So my guess about the meaning of ansible.pwd was wrong and I don't see any other fact.  take it out. Sent from my BlackBerry® PlayBook™www.blackberry.comFrom: "Jerry Vonau" <notifications@github.com>To: "XSCE/xsce" <xsce@noreply.github.com>CC: "Tim Moody" <tim@timmoody.com>Sent: 8 February, 2015 6:30 PMSubject: Re: [xsce] REVIEW - run git anywhere (but don't move it afterwards) (#208)when running from the GUI console:



failed: [127.0.0.1] => {"failed": true, "item": ""}
msg: Configuration directory (/opt/schoolserver/xsce)does not match actual directory (/opt/schoolserver/xsce_cmdsrv). Please consult documentation. Exiting.

PR #209, opened 2015-02-06
copy git repo if ansible is not run from 'xsce_dir'

jvonau:


tim-moody:
So if this is the first time through, the install will happen from the non-xsce_dir location, but will copy the repo where everything expects it to be.

Subsequently, even if you run from the non-xsce_dir location, sourcing xsce.env will put you into the correct one.

A couple of comments:

1) I have not had good luck with synchronize.  I’m not even sure it exists in 1.4 which is on f18.  So a plain old cp might be better or even a git clone as any subsequent pulls should go to the new location.  We also need to tell the user somehow so he or she doesn’t pull to the old location.
2) I would put this as early as possible, into the location where I have the fail in PR208 as it is a fix rather than a fail.

Everything else in 208 should still stand.


From: Jerry Vonau 
Sent: Friday, February 06, 2015 3:27 PM
To: XSCE/xsce 
Subject: [xsce] copy git repo if ansible is not run from 'xsce_dir' (#209)


  https://github.com/XSCE/xsce/pull/209

Commit Summary
  a.. copy git repo if ansible is not run from 'xsce_dir' 
File Changes
  a.. M roles/1-prep/tasks/computed_vars.yml (4) 
Patch Links:
  a.. https://github.com/XSCE/xsce/pull/209.patch 
  b.. https://github.com/XSCE/xsce/pull/209.diff

georgejhunt:
Has this been tested? Does this work? Looks to me that the test is PWD against {{xsce_base}} and the synchronize is PWD against {{ xsce_dir }} -- and not the same level in the FS

tim-moody:
I haven’t tested it.  Can’t speak for Jerry.  I think it assumes that {{ xsce_dir }}  = {{xsce_base}}/xsce, which is currently true, but not required to be.  I assume the synchronize command takes source dir as src and parent of target dir as dest.

My concern is more making sure the user is aware of the dipsy doodle and doesn’t continue to use the old location of the git repo to pull changes and then wonders why they never get applied.

This problem only arises if someone moves things after the offline install or doesn’t follow the documentation for the online install.

jvonau:
On Fri, Feb 6, 2015 at 3:39 PM, Tim Moody <notifications@github.com> wrote:

> I haven’t tested it. Can’t speak for Jerry. I think it assumes that {{
> xsce_dir }} = {{xsce_base}}/xsce, which is currently true, but not required
> to be.


well sort of:

xsce_base: /opt/schoolserver
xsce_dir: "{{ xsce_base }}/xsce"

And the instructions are to use /opt/schoolserver/xsce, I was just going
with the flow.



> I assume the synchronize command takes source dir as src and parent of
> target dir as dest.
>
>
yes works with ansible 1.4 http://docs.ansible.com/synchronize_module.html



> My concern is more making sure the user is aware of the dipsy doodle and
> doesn’t continue to use the old location of the git repo to pull changes
> and then wonders why they never get applied.
>
>
I agree, I would expect a user to git pull in /opt/schoolserver/xsce when
online to update. In the future I can see a git pull within the dipsy
doodle when that is available but that falls flat when offline.



> This problem only arises if someone moves things after the offline install
> or doesn’t follow the documentation for the online install.
>
>
Yea I could call this at the same point as PR208  but it's a moot point
now, with "cd $XSCE_DIR" in all the shell script ansible_env.PWD will
always return XSCE_DIR when not on the first run so there is no chance of
self updating now when run from a non-normal location.




> From: georgejhunt
> Sent: Friday, February 06, 2015 4:12 PM
> To: XSCE/xsce
> Cc: Tim Moody
> Subject: Re: [xsce] copy git repo if ansible is not run from 'xsce_dir'
> (#209)
>
>
> Has this been tested? Does this work? Looks to me that the test is PWD
> against {{xsce_base}} and the synchronize is PWD against {{ xsce_dir }} --
> and not the same level in the FS
>
>
from: http://www.comentum.com/rsync.html

When using "/" at the end of source, rsync will copy the content of the
last folder.
When not using "/" at the end of source, rsync will copy the last folder
and the content of the folder.

When using "/" at the end of destination, rsync will paste the data inside
the last folder.
When not using "/" at the end of destination, rsync will create a folder
with the last destination folder name and paste the data inside that
folder.

PR #210, opened 2015-02-07
Network with no gateway.

jvonau:
Without a gateway present ansible crashes hard at:
TASK: [network | Discover LAN iface] ****************************************** 
fatal: [127.0.0.1] => One or more undefined variables: 'dict object' has no attribute 'alias'

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/root/xsce-base.retry

127.0.0.1                  : ok=42   changed=7    unreachable=1    failed=0   

Patches 1 & 2 reorder the setting of xsce_lan_iface to avoid the running of shell code if xsce_lan_iface declared in local_vars.
Patch 3 allows the skipping of the of unneeded gateway play-books, and disables iptables when none is declared for xsce_wan_iface  in local_vars. The disabling of the services is not addressed, just bypassing those playbooks to run to success.  Those unneeded services can be disabled in the GUI later if needed

### DIFF
--- a/roles/gateway/tasks/dhcpd.yml
+++ b/roles/gateway/tasks/dhcpd.yml
@@ -31,3 +31,11 @@
   service: name=dhcpd 
            enabled=yes
            state=restarted
+  when: xsce_lan_iface != ""
+
+- name: Disable dhcpd service
+  service: name=dhcpd 
+           enabled=no
+           state=stopped
+  when: xsce_lan_iface == ""
+
--- a/roles/gateway/tasks/main.yml
+++ b/roles/gateway/tasks/main.yml
@@ -47,7 +47,6 @@
     - gateway
 
 - include: wondershaper.yml
-  when: xsce_lan_iface != "" and xsce_wan_iface != "none"
   tags:
     - wondershaper
     - gateway
--- a/roles/gateway/tasks/main.yml
+++ b/roles/gateway/tasks/main.yml
@@ -42,7 +42,6 @@
     - gateway
 
 - include: squid.yml
-  when: xsce_lan_iface != "" and xsce_wan_iface != "none"
   tags:
     - squid
     - gateway
--- a/roles/gateway/tasks/main.yml
+++ b/roles/gateway/tasks/main.yml
@@ -37,7 +37,6 @@
   file: dest=/etc/sysconfig/xs_httpcache_on state=absent
 
 - include: dhcpd.yml
-  when: xsce_lan_iface != ""
   tags:
     - dhcpd
     - gateway
--- a/roles/gateway/tasks/main.yml
+++ b/roles/gateway/tasks/main.yml
@@ -33,30 +33,41 @@
     - { src: "gateway/iptables.service" , dest: "/etc/systemd/system/iptables.service" }
     - { src: "gateway/xs-gen-iptables" , dest: "/usr/bin/xs-gen-iptables" }
 
+- name: delete proxy flag
+  file: dest=/etc/sysconfig/xs_httpcache_on state=absent
+
 - include: dhcpd.yml
-  when: 'xsce_lan_iface != ""' 
+  when: xsce_lan_iface != ""
   tags:
     - dhcpd
     - gateway
 
 - include: squid.yml
-  when: 'xsce_lan_iface != ""'
+  when: xsce_lan_iface != "" and xsce_wan_iface != "none"
   tags:
     - squid
     - gateway
 
 - include: wondershaper.yml
-  when: 'xsce_lan_iface != ""'
+  when: xsce_lan_iface != "" and xsce_wan_iface != "none"
   tags:
     - wondershaper
     - gateway
 
 - name: Run iptables 
   command: /etc/sysconfig/iptables-config
+  when: xsce_wan_iface != "none"
 
 - name: Enable iptables service 
   service: name=iptables
            enabled=yes
+  when: xsce_wan_iface != "none"
+
+- name: Disable iptables service 
+  service: name=iptables
+           enabled=no
+           state=stopped
+  when: xsce_wan_iface == "none"
 
 - name: Create gateway flag
   shell: echo 1 > /etc/sysconfig/olpc-scripts/setup.d/installed/gateway
--- a/roles/gateway/tasks/main.yml
+++ b/roles/gateway/tasks/main.yml
@@ -54,18 +54,15 @@
 
 - name: Run iptables 
   command: /etc/sysconfig/iptables-config
-  when: xsce_wan_iface != "none"
 
 - name: Enable iptables service 
   service: name=iptables
            enabled=yes
-  when: xsce_wan_iface != "none"
 
 - name: Disable iptables service 
   service: name=iptables
            enabled=no
            state=stopped
-  when: xsce_wan_iface == "none"
 
 - name: Create gateway flag
   shell: echo 1 > /etc/sysconfig/olpc-scripts/setup.d/installed/gateway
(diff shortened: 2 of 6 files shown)


tim-moody:
This is due to missing }, fixed in 208.
From: Jerry Vonau 
Sent: Saturday, February 07, 2015 7:10 AM
To: XSCE/xsce 
Subject: [xsce] Network with no gateway. (#210)

Without a gateway present ansible crashes hard at:
TASK: [network | Discover LAN iface] ****************************************** 
fatal: [127.0.0.1] => One or more undefined variables: 'dict object' has no attribute 'alias'

PLAY RECAP ******************************************************************** 
to retry, use: --limit @/root/xsce-base.retry

127.0.0.1 : ok=42 changed=7 unreachable=1 failed=0 

Patches 1 & 2 reorder the setting of xsce_lan_iface to avoid the running of shell code if xsce_lan_iface declared in local_vars.
Patch 3 allows the skipping of the of unneeded gateway play-books, and disables iptables when none is declared for xsce_wan_iface in local_vars. The disabling of the services is not addressed, just bypassing those playbooks to run to success. Those unneeded services can be disabled in the GUI later if needed


  https://github.com/XSCE/xsce/pull/210

Commit Summary
  a.. run detection only when set to auto, local_vars knows best 
  b.. move condition 
  c.. respect none as a xsce_wan_iface if set in vars 
File Changes
  a.. M roles/gateway/tasks/main.yml (17) 
  b.. M roles/network/tasks/main.yml (33) 
Patch Links:
  a.. https://github.com/XSCE/xsce/pull/210.patch 
  b.. https://github.com/XSCE/xsce/pull/210.diff

georgejhunt:
For me, there is merit in {{ xsce_wan_iface }} = none. Before we had to look at lan_iface == auto to infer something about wan.  

I ran into the missing ansible.if.alias, but had not traced it back to missing "}". And this PR disables iptables, when it is an appliance, which Tim and I had some discussion about.

I was hoping to ask you Jerry, to take on this whole area or thinking through, testing, the transition between appliance and gateway, back and forth.  Maybe this PR does it all, or maybe there will be additional work, testing, failures, etc

Nice to have you on board again Jerry

tim-moody:
Please see if 208 fixes the problem of the hard halt, in particular https://github.com/XSCE/xsce/commits/master/roles/1-prep/templates/local_facts.fact.j2

I agree there is a bigger question of cases that we don’t handle.

We currently support two nics, with one as upstream and one as downstream (‘gateway’).  We also support one nic as upstream (‘appliance’).

There is also the case of one nic as downstream, but you can not do an online install in this configuration because you are not online.

But after a gateway install, pulling a nic will not put you in appliance, because of iptables (going the other way works).  You might also do an appliance install, but want to then take it somewhere else where it should serve as dhcpd, dns, etc. server with no upstream connection.

I had hoped to handle all these cases in the console and had even contemplated a series of network diagrams that the user could select from, but the underlying ansible code is not there either.

jvonau:
Sorry about #208 but not fixed, this error occurs in the network playbook
as reported.

./runtags network

PLAY [all]
********************************************************************

GATHERING FACTS
***************************************************************
ok: [127.0.0.1]

TASK: [network | Discover LAN iface]
******************************************
fatal: [127.0.0.1] => One or more undefined variables: 'dict object' has no
attribute 'alias'

PLAY RECAP
********************************************************************
           to retry, use: --limit @/root/xsce.retry

127.0.0.1                  : ok=1    changed=0    unreachable=1
failed=0

because there is no gateway for ansible_default_ipv4.alias to detect, you
need something to suppress the detection that should only be run when set
to 'auto'.

Easy to reproduce just "route delete default" before you run ansible.

jvonau:
The rest of the gateway services with logic to enable/disable the services now in the playbooks.

tim-moody:
I'll give this a try, but I can't until Tuesday.Sent from my BlackBerry® PlayBook™www.blackberry.comFrom: "Jerry Vonau" <notifications@github.com>To: "XSCE/xsce" <xsce@noreply.github.com>CC: "Tim Moody" <tim@timmoody.com>Sent: 7 February, 2015 9:39 PMSubject: Re: [xsce] Network with no gateway. (#210)The rest of the gateway services with logic to enable/disable the services now in the playbooks.

—Reply to this email directly or view it on GitHub.

  
    
    
  
  

Sent from my BlackBerry® PlayBook™www.blackberry.com

tim-moody:
I’m still not clear on how to reproduce the problem that this PR addresses.  Does this only relate to offline install?  If I don’t have a gateway, where do I get the rpms, etc. to do the install.

I do the following:

on a vm with two nics (should this be one?)
install mininmal f20
clone xsce master to /opt/schoolserver and cd xsce
./install-console

am I correct so far?
what next? remove or disconnect a nic?  change a flag? under what circumstances would someone issue route delete default?

tim-moody:
tested PR 210

./install-console

TASK: [1-prep | Add location section to config file] **************************
changed: [127.0.0.1] => (item={'option': 'xsce_base', 'value': u'/opt/schoolserver'})
changed: [127.0.0.1] => (item={'option': 'xsce_dir', 'value': u'/opt/schoolserver/xsce'})

TASK: [1-prep | set_fact ] ****************************************************
fatal: [127.0.0.1] => One or more undefined variables: 'unicode object' has no attribute 'phplib_dir'

PLAY RECAP ********************************************************************
           to retry, use: --limit @/root/xsce-base.retry

127.0.0.1                  : ok=12   changed=10   unreachable=1    failed=0

This is due to missing } in local_facts.fact.j2 and was fixed in 208.

jvonau:
Once you have your install completed, you should be able to run turn off
services that are installed, from within the GUI or with runtags. You can't
to do that now when there is no gateway to detect. To reproduce disconnect
the wan nic or 'route delete default.' before using the GUI or runtags.

jvonau:
So you should not have that error then, there is nothing in this patch that
touches 1-prep.

tim-moody:
this was broken in 202 and fixed in 208. (https://github.com/jvonau/xsce/commit/49e1dd486627d57853f7324eb3d7c8d75fce56b0)

I have reproduced the error reported for 210 in both runtags and console.

I think the thing to do is merge 212, which fixes the console problem reported in 208:
when running from the GUI console:


failed: [127.0.0.1] => {"failed": true, "item": ""}
msg: Configuration directory (/opt/schoolserver/xsce)does not match actual directory (/opt/schoolserver/xsce_cmdsrv). Please consult documentation. Exiting.

and then rebase 210 on 212.  Then it can be tested.

jvonau:
Seems reasonable.

tim-moody:
while we’re at it, any idea why default_vars.yml 

xsce_wan_iface: "{{ ansible_default_ipv4.interface }}"

doesn’t break?

jvonau:
I was wondering about that too, I'll side with "if not present return
blank" when gathering facts.

tim-moody:
if there’s a default it returns

"ansible_default_ipv4": {"address": "192.168.0.31", "alias": "p2p1", "gateway": "192.168.0.1", "interface": "p2p1", "macaddress": "08:00:27:ec:93:35", "mtu": 1500, "netmask": "255.255.255.0", "network": "192.168.0.0", "type": "ether"}

if not it returns

"ansible_default_ipv4": {}

so either both alias and interface or neither.  seems funny.  hope they don’t change it at some point.

jvonau:
The reason why 'ansible_default_ipv4.interface' didn't fail is because 'ansible_default_ipv4.alias' was accessed before using what was available in 'xsce_wan_iface' at that point in time. When moving the ordering of the interfaces you run into 'ansible_default_ipv4.interface' popping up the same type of error as reported. Closed in favour of #218.

PR #213, opened 2015-02-10
REVIEW - Networking Design Discussion

tim-moody:
It is time to revisit the networking modes supported by xsce and which services each supplies, how they are configured, and in what ways it is possible to change from one to another.

Modes
---------

* Appliance (one nic with an upstream connection: member of a larger wan/lan)
* Gateway (two nics, one with an upstream connection and one with a downstream connection
     to XO or other client machines)
* Network Controller (one nic, possibly connected upstream during online installation and later
     changed to downstream)
* Others?

Flags
-------

What are the possible values of the following and what do they do?

* xsce_wan_iface:
* xsce_lan_iface:
* xsce_gateway_enabled
* ansible_default_ipv4.interface
* Others?

To switch from one mode to another which flags are changed and to what values?

State
-------

State is stored in various locations.  What is the relationship between them?

* /etc/xsce/xsce.ini
* /etc/sysconfig
* /etc/sysconfig/olpc-scripts/setup.d/
* /etc/sysconfig/olpc-scripts/setup.d/installed/

Networking is discussed in contradictory ways in a number of documents, which should be revised.

There are probably other issues.

### DIFF
--- a/docs/NETWORKING.rst
+++ b/docs/NETWORKING.rst
@@ -1,6 +1,8 @@
 Networking
 ==========
 
+THIS DOCUMENT AND THE NETWORKING SECTION OF INSTALL AND OTHER DOCS NEEDS REVISING.
+==================================================================================
 
 The schoolserver community edition has two modes of operation:
 


jvonau:
On Tue, Feb 10, 2015 at 10:31 AM, Tim Moody <notifications@github.com>
wrote:

> It is time to revisit the networking modes supported by xsce and which
> services each supplies, how they are configured, and in what ways it is
> possible to change from one to another.
>
both from the console and cli.

> Modes
>
>    - Appliance (one nic with an upstream connection: member of a larger
>    wan/lan)
>
> no dhcpd, caching disabled, iptables disabled.

>
>    - Gateway (two nics, one with an upstream connection and one with a
>    downstream connection to XO or other client machines)
>
> dhcpd active, caching optional, iptables active.

>
>    - Network Controller (one nic, possibly connected upstream during
>    online installation and later changed to downstream)
>
>  Network Controller, I like that definition.
dhcpd active, caching disabled, iptables disabled.

>
>    - Others?
>
> Think the above three cover the use base. The disabling of the caching is
more for resource saving where there is no benefit from having the services
running.

>
>
> Flags
>
> What are the possible values of the following and what do they do?
>
>    - xsce_wan_iface:
>    - xsce_lan_iface:
>    - xsce_gateway_enabled
>    - ansible_default_ipv4.interface
>    - Others?
>
> To switch from one mode to another which flags are changed and to what
> values?
>

Think we need to discuss the current behaviour.

> State
>
> State is stored in various locations. What is the relationship between
> them?
>
>    - /etc/xsce/xsce.ini
>    - /etc/sysconfig
>    - /etc/sysconfig/olpc-scripts/setup.d/
>    - /etc/sysconfig/olpc-scripts/setup.d/installed/
>
> Networking is discussed in contradictory ways in a number of documents,
> which should be revised.
>
> There are probably other issues.
>
Off the top of my head, wondershaper needs the default_vars treatment.

tim-moody:
So, could we have a variable xsce_network_mode (or role),  calculated in computed_vars based on all the variables and facts, which are known by then, and then use that for conditions, rather than a list of interface = “” or not = or enabled, etc.?  That way all the logic could be handled in one place, rather than having to fix up each condition when the understanding of auto or none or enabled changes.  So,

xsce_network_mode: “Appliance”
when:

xsce_network_mode: “Gateway”
when:

xsce_network_mode: “NetController” (or “LanController”)
when:

We need to take into consideration what was discovered and what the user asked for and decide what to do when the user asked for something that can not be accommodated, like a gateway with one nic.)

also add named to the below.
by caching we mean caching and filtering.

>
>> Modes
>>
>> - Appliance (one nic with an upstream connection: member of a larger
>> wan/lan)
>>
>> no dhcpd, caching disabled, iptables disabled.
no named
>
>>
>> - Gateway (two nics, one with an upstream connection and one with a
>> downstream connection to XO or other client machines)
>>
>> dhcpd active, caching optional, iptables active.
named active
>
>>
>> - Network Controller (one nic, possibly connected upstream during
>> online installation and later changed to downstream)
>>
>> Network Controller, I like that definition.
>dhcpd active, caching disabled, iptables disabled.
named active
>
>>
>> - Others?
>>
>> Think the above three cover the use base. The disabling of the caching is
>more for resource saving where there is no benefit from having the services
>running.

jvonau:
Yea think that is possible, you are just looking to use (xsce_network_mode)
where I used (xsce_wan_iface != "none" and xsce_lan_iface != "") for the
services. Or are you thinking of using xsce_network_mode to "set_fact
squid_enabled True/False" similar to the handling of mysqld'? If so then I
think dhcpd, wondershaper needs some work in default_vars.

tim-moody:
One of the problems we have is that we don’t distinguish well between directives, squid_enabled or gateway_enabled, which means I want squid enabled, etc/, vs. state, such as mysql_service, which means in this environment the proper package to install for mysql.  The former is used for setting the configuration and the latter is a value that can’t be set by the end user.

I mean xsce_network_mode to be the latter.  There is gateway_enabled, which is the former and means I want the gateway to be enabled (if it can).  I take squid_enabled to be a configuration variable, but I like your idea that we suppress True to False if enabling squid can’t be done or makes no sense.  I wouldn’t change the configuration file, though.  It raises another question as does the gateway_enabled example; do we ignore things that don’t make sense or do they constitute an error.  I’m inclined towards ignoring.

I think it would be great if any logic that turns things on or off, beyond simple flags set by the user, were calculated in computed_vars.

tim-moody:
I’m wondering how we tell whether we are in appliance or netcontroller mode.  One answer would be if I only have one nic, then if there is a default ipv4 I’m in appliance, else I’m in netcontroller.  This can get tricky if the machine has a usb dongle and they got switched so the mac address is wrong.  Will linux handle the changed mac address problem for dhcp configured interfaces?  The other way mac addresses can be wrong is offline install, where we simply untar the root file system from an install on a different machine.

But if there are two nics, then I think we need a netcontroller_enabled flag along side the gateway_enabled flag, both of which are user settable.

But we also need to handle the case where there is one nic and the server was in appliance mode when it was set up, because it needed an internet connection for rpms or something, but it gets deployed in netcontroller mode.

tim-moody:
Another aspect to networking that needs work is discovery of the server.  In a gateway, the server provides dns and assuming you didn’t override the host name, you should be able to connect to ‘schoolserver’.  But I wish we advertised sshd and httpd on avahi as schoolserver, so that even if the host name is changed it can still be found.  I also think that some later versions of XO OS don’t find schoolserver on dns.  In appliance mode this is even more important as we don’t know what other machines are on the network or what is providing dns.

tim-moody:
Regarding Appliance vs. NetController, we could also say that if there is only one nic then we should only supply dhcpd services if nothing else is.  This sounds more like something to consider at boot time rather than install/configure time.

jvonau:
There maybe an issue with ifcfg- files, we only create them, but never cleaning up if no longer valid.

georgejhunt:
As an aside, when loading a new OS from USB stick, I found that ansible did
a pretty good job of self-configuring, if I deleted the ifcfg files, and
let it start over.  Would it work to delete them as the first step of an
ansible playbook for the network role?

jvonau:
Deleting is not a bad idea, but which ones? I'd hate to delete an ifcfg
file that is used in a static ip config for WAN.

Think in place of deleting we could just use the same name for the ifcfg-
file, the contents gets replaced on every run and we don't care what the
file is named just the contents, for device and macaddress.

-            dest=/etc/sysconfig/network-scripts/ifcfg-{{xsce_lan_iface}}
+            dest=/etc/sysconfig/network-scripts/ifcfg-LAN

Thoughts?

georgejhunt:
Seems like a good approach. Though to take care of the situation where the
console interaction requests transition from gateway to appliance, I think
the delete WAN would be required.

In the situation you point to (staticly defined WAN), will not the template
file be recreated later in the playbook?

tim-moody:
I like logical names if it works.  Do you get a problem is someone switches networking mode?

tim-moody:
The other tricky thing about deleting is I think NM creates some of these files.

jvonau:
Perhaps combining network/gateway with the meta trick might be the way to go given the closely coupled integration? With that in place doing dnsmasq as an option/replacement become much easier in the future.

tim-moody:
I could see having a single role for network and gateway as the gateway is really just one possible network configuration.  I think meta is too granular.

tim-moody:
Hosts file: need to think about what to do if no lan iface.  should wan ip go there?  what if not static?

as per PR 221.

jvonau:
What should the firewall state be while not in Gateway mode? We currently fake out iptables when one interface is available by recording the same interface for both WAN and LAN. The end effect is that all traffic on that interface is allowed based on the iptable rule "$IPTABLES -A INPUT -m state --state NEW -i  $lan -j ACCEPT", that is effectively wide open. I would prefer not have iptables rules loaded when none are really needed. On a side note, should the ports needed for the GUI optionally be opened for the WAN?

tim-moody:
good question.  I think open is probably OK for now, but TK wanted to expose a VM to the open internet with an xsce in appliance mode and that makes me nervous with no firewall.  by the same token we would open the ports to all our apps for an appliance, so we are only protecting against vulnerabilities on ports we don’t use.

jvonau:
Tim mentioned hosts file when in appliance mode, have a look at https://github.com/XSCE/xsce/commit/2e5a65b25a0cdd60e7a32d5718618929b59cd60b. currently untested. Think we can close this PR once #239 is committed.

tim-moody:
Closing due in light of networking changes now merged.

PR #214, opened 2015-02-12
RFC Network3

jvonau:
This is a RFC in regards to the Networking Design Discussion in #213. Should be feature complete.


tim-moody:
This is a huge step in the right direction imho.

Comments/Questions:

 5714851 - this is something of a conflict between logical and physical organization, could we have the best of both by keeping at least the Filters heading above dansguardian, etc.

630090b, 480b6eb - I assume there are no interactions between config scripts and simply turning off services.  Does squid change port forwarding, for example?

I assume that adding the default for named, etc. to enabled means if you can.

 c2e98ab -I know this has been there quite awhile, but I never understood it.

 set_fact:
    xsce_wan_iface: {{ xsce_wan_iface }}
   when: 'xsce_wan_iface != "auto" or xsce_wan_iface == "none" '

Isn't that like

If A = "auto" or A = "none" then
   A = A
else
   A = A

what values exactly can xsce_wan_iface have?

 0e93c6d - does this work?  didn't know you could set multiple facts with one statement.

with these changes computed_vars has become large.  you could also include a network_vars or some such.

both xsce_wan_iface and xsce_lan_iface can be auto or none?  Both feel to me like something that should come from the environment, not a configuration setting.  At the very least we need a truth table that indicates their settings and the resulting networking configuration.

typo 
name: Add location [network] section to config file
+  ini_file: dest='{{ xsce_config_file }}'
+            section=network
+            option='{{ item.option }}'
+            value='{{ item.value }}'

 f3b8894 - it looks like gateway runs after network. how does that work?

ec6cc91 - glad to see this.  hope it works and there are no scripts depending on it.

4405543 - is the point of A = A to get the quotes for comparison?

6902304 - this looks done and undone?  is reread facts still near the creation of local facts?

0474a49 - I find booleans not always clear in this environment, for example I think True does not equal true.  I follow the practice of using True and False.

jvonau:
Nothing is written in stone yet. lol

On Thu, Feb 12, 2015 at 3:11 PM, Tim Moody <notifications@github.com> wrote:

> This is a huge step in the right direction imho.
>
Thanks, we can cherry-pick ideas.


> Comments/Questions:
>
> 5714851
> <https://github.com/XSCE/xsce/commit/57148519ae73bd7ad35b0fd9d3f11d5ffbab8145>
> - this is something of a conflict between logical and physical
> organization, could we have the best of both by keeping at least the
> Filters heading above dansguardian, etc.
>
Yea think those might be better placed there, was just trying to keep
services straight in my head.


> 630090b
> <https://github.com/XSCE/xsce/commit/630090be217db9b0e95db04ad9f8db29b6f10093>,
> 480b6eb
> <https://github.com/XSCE/xsce/commit/480b6eb9fce135a5d0db6c85274ec1a91a397e42>
> - I assume there are no interactions between config scripts and simply
> turning off services. Does squid change port forwarding, for example?
>
Good point, the xs_httpcache_on should really be part of turning on/off the
service.

> c2e98ab
> <https://github.com/XSCE/xsce/commit/c2e98ab24ad85dc22b68887903e32b82e5bd7287>
> -I know this has been there quite awhile, but I never understood it.
>
Just working with what is currently present.

> set_fact:
> xsce_wan_iface: {{ xsce_wan_iface }}
> when: 'xsce_wan_iface != "auto" or xsce_wan_iface == "none" '
>
> Isn't that like
>
> If A = "auto" or A = "none" then
>
No, if A is "auto" or A is not "none" then
pass A

> A = A
> else
> A = A
>
what values exactly can xsce_wan_iface have?
>
I took it to mean values passed from vars/*, where it looked like you could
pass, based on the code, auto/none/value for xsce_lan_iface. Just wanted to
be able to have an override available in local_facts.



> 0e93c6d
> <https://github.com/XSCE/xsce/commit/0e93c6df2a40dae5fe9fb701dfb2ce17488decb6>
> - does this work? didn't know you could set multiple facts with one
> statement.
>
Seems to work great.


> with these changes computed_vars has become large. you could also include
> a network_vars or some such.
>
I was toying with using an include statement, whichever way you prefer.


> both xsce_wan_iface and xsce_lan_iface can be auto or none? Both feel to
> me like something that should come from the environment, not a
> configuration setting. At the very least we need a truth table that
> indicates their settings and the resulting networking configuration.
>
I'm thinking local_vars:
auto = detect
none =  don't detect this interface
value = use what is put here when not none.

Think we need to do something in git about local_vars.yml getting
overwritten on a git pull.

> typo
> name: Add location [network] section to config file
>
>    - ini_file: dest='{{ xsce_config_file }}'
>    - section=network
>    - option='{{ item.option }}'
>    -
>
>           value='{{ item.value }}'
>
>
> Yea, didn't change the name line, that can be anything you want. Was
looking to record the computed network layout.


>    -
>
>     f3b8894
>    <https://github.com/XSCE/xsce/commit/f3b8894fabab919616e46d16c852b972450ed872>
>    - it looks like gateway runs after network. how does that work?
>
> Other way around currently network is before gateway. Now gateway runs
before network, works ok providing xs_*an_device.j2 is in place before
gateway's playbook is run. The idea is to let NetworkManager's dispatcher
scripts (re)start the service when an interface appears, so the NM restart
is run last. This will allow not having to start a service as a playbook is
run, we can remove state=started/restarted for those services that have
scripts in /etc/NetworkManager/dispatcher.d/.

>
>    -
>
> ec6cc91
> <https://github.com/XSCE/xsce/commit/ec6cc91d199da2f84c805248eef1afb02e6b2cd6>
> - glad to see this. hope it works and there are no scripts depending on it.
>
Just xs-gen-iptables as far as I can recall. Well the logic was changed in
c2e98ab
<https://github.com/jvonau/xsce/commit/c2e98ab24ad85dc22b68887903e32b82e5bd7287>.
Now we are not lying about interfaces found. No interface no entry recorded.


> 4405543
> <https://github.com/XSCE/xsce/commit/44055439d1e92ed79000b56449532d9f07b4c215>
> - is the point of A = A to get the quotes for comparison?
>
More like to test if there are any entries returned. Not interested in what
the results are just if any are present, might need further work.

6902304
> <https://github.com/XSCE/xsce/commit/69023047b24114b9e77721e44f38b796147c0b30>
> - this looks done and undone? is reread facts still near the creation of
> local facts?
>
re-read facts didn't move, the file copying is now in prep.yml and that is
run before computed_vars.yml according to main.yml.


> 0474a49
> <https://github.com/XSCE/xsce/commit/0474a49b722a4c2987d88a67f2f5a705e94f32be>
> - I find booleans not always clear in this environment, for example I think
> True does not equal true. I follow the practice of using True and False.
>
I think True might equal true else the *_enabled booleans should fail. Or
is there other logic involved?

tim-moody:
local facts does it as soon as you run ansible, so if you did an edit and another pull before running you could have a problem, but not too likely except maybe for developers.
I put my local_vars in github in tim-moody/xsce-local along with any other customizations, such as keys.
# while we're here untrack local vars
git update-index --assume-unchanged vars/local_vars.yml

>Think we need to do something in git about local_vars.yml getting
>overwritten on a git pull.

tim-moody:
>> f3b8894
>> <https://github.com/XSCE/xsce/commit/f3b8894fabab919616e46d16c852b972450ed872>
>> - it looks like gateway runs after network. how does that work?
>>
>> Other way around currently network is before gateway. Now gateway runs
>before network, works ok providing xs_*an_device.j2 is in place before
>gateway's playbook is run. The idea is to let NetworkManager's dispatcher
>scripts (re)start the service when an interface appears, so the NM restart
>is run last. This will allow not having to start a service as a playbook is
>run, we can remove state=started/restarted for those services that have
>scripts in /etc/NetworkManager/dispatcher.d/.

sorry, I misspoke, network used to run first and I thought gateway needed that.  it’s a little counterintuitive to me that you can have the gateway working without networking.

some services might need to be restarted as the playbook runs because that is needed for subsequent steps.  I had substantial problems with NM dispatcher as my comments indicate, but maybe you can get it to work.

I guess these are OK unless squid needs named.  I hope this doesn’t mean a reboot as it doesn’t always happen.

-rwxr-xr-x. 1 root root 175 Jan 14  2014 00-netreport
-rwxr-xr-x. 1 root root 933 Aug 14 09:07 11-dhclient
-rwxr-xr-x. 1 root root 597 Aug 14 09:07 12-dhcpd
-rwxr-xr-x. 1 root root 344 Dec 10 13:51 13-named
-rwxr-xr-x. 1 root root 287 Aug 15 06:15 20-chrony
-rwxr-xr-x. 1 root root 138 Sep 12 06:59 20-squid

jvonau:
It's not a problem if you never do a git pull in xsce_dir, but that is we
are telling users to place their custom stuff in. I was hoping that we
could add optionally updating a base install with the latest bits using git.

I was thinking maybe placing a copy of local_vars or another file into
/etc/ansible/facts.d for passwords, ssh_keys that are not meant to be
changed in the field after the first install, that will help keep settings
local to a machine for future use.

Just some ideas

jvonau:
Those are run in order, denoted by the digits. I now see that named is
doing a reload not restart like squid does so named has to be started from
the playbook.

tim-moody:
It’s currently not a problem if you do a pull into xsce_dir unless you do it between editing local_vars and first run; that pull will fail because of uncommitted changes so your setting will not be lost.  Any pulls after that will not cause a problem, because tracking is stopped on the first run and local_vars remains as the local version of the overridden variables.  I tried putting the file somewhere else, but then the first run fails because the file does not yet exist and must be included in the top level playbook.  I don’t mind if there is a copy of local_vars in /etc/xsce, but it is tricky to orchestrate a file that must be present before any code is run.  In any event the mechanism would be the same as now – you would have to do a first run before things are in place.

jvonau:
Think I worked around the error reported in #210 by using "is defined" with "register" bypassed the error first reported, and adjusting the logic allows to auto detect both LAN and WAN when set to "auto"

tim-moody:
hadn’t read this yet when I commented on 218, but thinking in the same direction.

jvonau:
Closing in light of #222

PR #215, opened 2015-02-12
BUGFIX gateway services

jvonau:
While working on the network revamp I noticed some of the gateway services don't have *_enabled stanzas, leaving no future way of disabling of these service from within ansible. These are optionally installed and configured currently only at install time. Remove the run-time LAN condition allowing playbooks to be accessed anytime, this allows the services to respect any *_enabled variables passed. This will allow installation of gateway services while not having a LAN present at installation time.

### DIFF
--- a/roles/gateway/tasks/dhcpd.yml
+++ b/roles/gateway/tasks/dhcpd.yml
@@ -27,7 +27,14 @@
         mode=0644
         state=file
 
+- name: Disable dhcpd service
+  service: name=dhcpd 
+           enabled=no
+           state=stopped
+  when: not dhcpd_enabled
+
 - name: Enable dhcpd service
   service: name=dhcpd 
            enabled=yes
            state=restarted
+  when: dhcpd_enabled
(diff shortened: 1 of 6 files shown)


tim-moody:
Are these the same as in the last PR?

tim-moody:
The ability to install for a future environment is good, but tricky.  If I do an online install I need to connect on the wan port, which in turn needs to connect to the internet.  Later, if the server is a gateway, I will not be able to connect to the console on that interface and will need to switch to the lan.  We need to make sure you can't get locked out.

jvonau:
Your concerns about the future are well founded but outside of the scope of
the PR. This bug fix address the fact that if you install with just WAN(no
LAN) you will not install any gateway services. This PR removes that
restriction with 95ca52c
<https://github.com/jvonau/xsce/commit/95ca52cad9fec954414a2aca7ab7ef4b73e77b42>.
The rest allow for disabling these services from local_facts which doesn't
exist at this point in time.

tim-moody:
I tested this on a gateway and an appliance with each of these first true and then false.  

dhcpd_enabled: False          
named_enabled: False          
dansguardian_enabled: False   
squid_enabled: False          
wondershaper_enabled: False   
xsce_gateway_enabled: False

it did what it advertises in that all these services were turned from on to off.

jvonau:
That is all this PR deals with, the inability to disable gateway services.

PR #216, opened 2015-02-13
Owncloud mysql

tim-moody:
move owncloud to mysql from sqlite

tested on 64 bit and xo4


tim-moody:
"overwrite.cli.url" => "http://{{ xsce_hostname }}.{{ xsce_domain }}/owncloud", still gets overwritten with the url from which the admin user and password are created, but we will leave it for the future.

PR #217, opened 2015-02-14
BUGFIX #210

jvonau:
Backport from #214 for #210 for better readability. grep -e doesn't like blanks or {} so use none instead then set to blank


jvonau:
closed in favour of #218

PR #220, opened 2015-02-19
Elgg 02 18

tim-moody:
upgrade elgg to 1.10.1
install from sample files
load database from dump


jvonau:
How about url={{ xsce_download_url }}/

PR #221, opened 2015-02-20
Selective gateway services.

jvonau:
Install all gateway services before detecting network. After detecting network set the gateway services based on network configuration.

### DIFF
--- a/roles/3-base-server/meta/main.yml
+++ b/roles/3-base-server/meta/main.yml
@@ -1,6 +1,6 @@
 dependencies:
-   - { role: gateway, tags: ['base','gateway'] }
    - { role: network, tags: ['base','network'] }
+   - { role: gateway, tags: ['base','gateway'] }
    - { role: httpd, tags: ['services','httpd','base'] }
    - { role: postgresql, tags: ['services','postgresql','base'] }
    - { role: authserver,  tags: ['services','authserver','base'] }   
--- a/roles/3-base-server/meta/main.yml
+++ b/roles/3-base-server/meta/main.yml
@@ -1,6 +1,6 @@
 dependencies:
-   - { role: network, tags: ['base','network'] }
    - { role: gateway, tags: ['base','gateway'] }
+   - { role: network, tags: ['base','network'] }
    - { role: httpd, tags: ['services','httpd','base'] }
    - { role: postgresql, tags: ['services','postgresql','base'] }
    - { role: authserver,  tags: ['services','authserver','base'] }   
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -1,3 +1,8 @@
+- include: named.yml
+  tags:
+    - named
+    - network
+
 # WAN
 - name: Set user configured wan fact
   set_fact:
@@ -154,8 +159,3 @@
            enabled=yes
            state=started
 - debug:  msg="hopefully now NM is restarted"
-
-- include: named.yml
-  tags:
-    - named
-    - network
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -75,16 +75,10 @@
     - services
     - squid
 
-- name: No LAN configured - Appliance mode
+- name: No LAN configured via vars
   set_fact:
-    xsce_lan_iface: "none"
-    dansguardian_enabled: False
-    squid_enabled: False
-    named_enabled: False
-    dhcpd_enabled: False
-    wondershaper_enabled: False
-    xsce_network_mode: "Appliance"
-  when: 'xsce_lan_iface == "none" or not discovered_lan is defined'
+    xsce_lan_iface: ""
+  when: xsce_lan_iface == "none"
   tags:
     - network-discover
     - gateway
@@ -97,6 +91,16 @@
     - services
     - squid
 
+- name: No LAN configured - Appliance mode
+  set_fact:
+    dansguardian_enabled: False
+    squid_enabled: False
+    named_enabled: False
+    dhcpd_enabled: False
+    wondershaper_enabled: False
+    xsce_network_mode: "Appliance"
+  when: 'xsce_lan_iface == "" and xsce_wan_iface != ""'
+
 - name: LAN configured - LanController mode
   set_fact:
     named_enabled: True
@@ -105,14 +109,14 @@
     squid_enabled: False
     wondershaper_enabled: False
     xsce_network_mode: "LanController"
-  when: 'xsce_lan_iface != "none" and xsce_wan_iface == "none"'
+  when: 'xsce_lan_iface != "" and xsce_wan_iface == "none"'
 
 - name: LAN configured - Gateway mode
   set_fact:
     named_enabled: True
     dhcpd_enabled: True
     xsce_network_mode: "Gateway"
-  when: 'xsce_lan_iface != "none" and xsce_wan_iface != "none"'
+  when: 'xsce_lan_iface != "" and xsce_wan_iface != "none"'
 
 - name: Add location section to config file
   ini_file: dest='{{ xsce_config_file }}'


tim-moody:
Some questions/comments:

if there's only one nic, is it wan or lan?  if it's an appliance it's connected upstream, if a lancontroller downstream.

how can we configure named before we know whether we even want it or what nic it should listen on?  is it included at the beginning and at the end?

I still find it confusing that xsce_lan_iface is sometimes state (none = none found) and sometimes a setting (none = don't use it). same for wan, and see first question.  I have read through this logic 10 times and I still wonder if it catches every case.

what will be the result (value of xsce_wan_iface) for the following

set_fact:
    xsce_wan_iface: "{{ xsce_wan_iface }}"
  when: 'xsce_wan_iface != "auto" or xsce_wan_iface != ""'

when xsce_wan_iface = "none" - "none"?
when xsce_wan_iface = "auto" - "auto"?
when xsce_wan_iface = 192.168.0.25 - 192.168.0.25?
when xsce_wan_iface = "" - ""?
when xsce_wan_iface = "p3p5" - "p3p5"?

it looks to me like  xsce_wan_iface always ends up "{{ xsce_wan_iface }}"

btw, after the next two declarations xsce_wan_iface can still be "auto" or anything besides none if ansible_default_ipv4.gateway is not defined.

once we get to the turning on and off of services this looks pretty good and I'm glad we have network mode in case any other role needs it.

I take it you don't explicitly set squid, etc. on for gateway because you want to respect anything from default or local_vars.

do we want 172.18.96.1 in hosts if this is an appliance?

could we rename ifcfg.j2 to lan-ifcfg.j2 in preparation for wan-ifcfg.j2 for static ip on wan in appliance or gateway?

jvonau:
On Fri, Feb 20, 2015 at 5:22 PM, Tim Moody <notifications@github.com> wrote:

> Some questions/comments:
>
> if there's only one nic, is it wan or lan? if it's an appliance it's
> connected upstream, if a lancontroller downstream.
>

This has not changed from the past, it's treated as WAN during install
time. Changing from a straight appliance install to a lancontroller type
deployment, as opposed to coming from gateway to lancontroller, was never
supported. Nothing changed there except all the gateway type services are
already installed just need configuring, no further rpms required, so at
least now you can use ansible from the cli. Adding an interface that was
not present during install would need cli (or ssh) access or ensure there
is nothing blocking the ports needed for the GUI to have the network
re-detected by ansible.

how can we configure named before we know whether we even want it or what
> nic it should listen on? is it included at the beginning and at the end?
>
Just wanted to have all the services installed before the network is
restarted. Makes using something else less of a pain in the future. Named
is bound to all interfaces by default that is why there is the dispatcher
script that reloads named as interfaces come and go.


> I still find it confusing that xsce_lan_iface is sometimes state (none =
> none found) and sometimes a setting (none = don't use it). same for wan,
> and see first question. I have read through this logic 10 times and I still
> wonder if it catches every case.
>

Just changed that because xsce_lan_iface can really be "" when detected as
such and discovered_lan is registered as soon as the shell code is run, it
is of no use to determine if there is no interface. see the last commit.


> what will be the result (value of xsce_wan_iface) for the following
>
> set_fact:
> xsce_wan_iface: "{{ xsce_wan_iface }}"
> when: 'xsce_wan_iface != "auto" or xsce_wan_iface != ""'
>
> when xsce_wan_iface = "none" - "none"?
> when xsce_wan_iface = "auto" - "auto"?
> when xsce_wan_iface = 192.168.0.25 - 192.168.0.25?
> when xsce_wan_iface = "" - ""?
> when xsce_wan_iface = "p3p5" - "p3p5"?
>
Only if you're messing with vars/*, trying to avoid the need for that and
still respect vars if present.


> it looks to me like xsce_wan_iface always ends up "{{ xsce_wan_iface }}"
>
Yes, when not set to auto or none in vars/* that is the point of being able
to use vars/.

btw, after the next two declarations xsce_wan_iface can still be "auto" or
> anything besides none if ansible_default_ipv4.gateway is not defined.
>
Can you clarify that for me please. xsce_wan_iface is set to "none" when 'not
ansible_default_ipv4.gateway is defined'


> once we get to the turning on and off of services this looks pretty good
> and I'm glad we have network mode in case any other role needs it.
>
> I take it you don't explicitly set squid, etc. on for gateway because you
> want to respect anything from default or local_vars.
>
Correct, need to respect the user's wishes.


> do we want 172.18.96.1 in hosts if this is an appliance?
>
Can we use #213 for this kind of question?  should we use the WAN ip with
appliance?

We could skip the host file treatment if LAN is absent, or fix it up.

> could we rename ifcfg.j2 to lan-ifcfg.j2 in preparation for wan-ifcfg.j2
> for static ip on wan in appliance or gateway?
>
I kind want to keep it in manageable chunks, that will be for a another PR

tim-moody:
I realize some of my comments are about legacy code, not new code, just thought it was an opportunity to clean up.  I’m OK with leaving things as long as they aren’t broken and work as at present.

To be clear, in default_vars or local_vars xsce_wan_iface and xsce_lan_iface can be set to

none – don’t use this interface even if you find it
auto – use whatever you find. can only use for wan if it is a gateway and for lan if it is not a gateway.  auto is the default for both.

can they also be explicitly set to a value, like p2p7?  I prefer not, because I don’t want to have to deal with mismatches between settings and what’s actually in the machine.

during the calculations these will be changed to

none – not used because so set or nothing useable found
value – the actual device to be used

I still think it would be more readable if we had xsce_wan_enabled and xsce_lan_enabled (and get rid of xsce_gateway_enabled)

False – don’t use this interface even if you find it
True – use whatever you find. can only use for wan if it is a gateway and for lan if it is not a gateway.  auto is the default for both.

and then assign none or an actual device to xsce_wan_iface and xsce_lan_iface based on enabled True/False and what is discovered.

> Named
> is bound to all interfaces by default that is why there is the dispatcher
>script that reloads named as interfaces come and go.

so none of the scripts relies on our interface variables?
>
>
>> I still find it confusing that xsce_lan_iface is sometimes state (none =
>> none found) and sometimes a setting (none = don't use it). same for wan,
>> and see first question. I have read through this logic 10 times and I still
>> wonder if it catches every case.
>>
>
>Just changed that because xsce_lan_iface can really be "" when detected as
>such and discovered_lan is registered as soon as the shell code is run, it
>is of no use to determine if there is no interface. see the last commit.
>
are you saying that shell: "/sbin/ip link show  | grep -v -e DOWN -e lo -e tun -e {{ ansible_default_ipv4.alias }}| grep mtu | gawk -F: '{print $2}' |  tr -d ' '  | head -n1 "

can return “” and that we accept it if it does?

>
>> what will be the result (value of xsce_wan_iface) for the following
>>
>> set_fact:
>> xsce_wan_iface: "{{ xsce_wan_iface }}"
>> when: 'xsce_wan_iface != "auto" or xsce_wan_iface != ""'
>>
>> when xsce_wan_iface = "none" - "none"?
>> when xsce_wan_iface = "auto" - "auto"?
>> when xsce_wan_iface = 192.168.0.25 - 192.168.0.25?
>> when xsce_wan_iface = "" - ""?
>> when xsce_wan_iface = "p3p5" - "p3p5"?
>>
>Only if you're messing with vars/*, trying to avoid the need for that and
>still respect vars if present.
>
>
>> it looks to me like xsce_wan_iface always ends up "{{ xsce_wan_iface }}"
>>
>Yes, when not set to auto or none in vars/* that is the point of being able
>to use vars/.

all I’m trying to say is this boils down to 

if (any condition) then A = A

so if the condition is met A is set to A and if not A remains A. It doesn’t do anything.

>
>btw, after the next two declarations xsce_wan_iface can still be "auto" or
>> anything besides none if ansible_default_ipv4.gateway is not defined.
>>
>Can you clarify that for me please. xsce_wan_iface is set to "none" when 'not
>ansible_default_ipv4.gateway is defined'
>
sorry, I think I took an or for an and.  so this really boils down to

set_fact: 
xsce_wan_iface: "none" 
when: not ansible_default_ipv4.gateway is defined' 

(no need to set to none if it is already none.)

jvonau:
On Sat, Feb 21, 2015 at 8:37 AM, Tim Moody <notifications@github.com> wrote:

> I realize some of my comments are about legacy code, not new code, just
> thought it was an opportunity to clean up. I’m OK with leaving things as
> long as they aren’t broken and work as at present.
>
> To be clear, in default_vars or local_vars xsce_wan_iface and
> xsce_lan_iface can be set to
>
> none – don’t use this interface even if you find it
> auto – use whatever you find. can only use for wan if it is a gateway and
> for lan if it is not a gateway. auto is the default for both.
>
>
OK, agreed


> can they also be explicitly set to a value, like p2p7? I prefer not,
> because I don’t want to have to deal with mismatches between settings and
> what’s actually in the machine.
>
>
OK, we could do a sanity check on other than "none auto" user provided
info. I would prefer not to, lets think of that as expert mode, use at your
own discretion.


> during the calculations these will be changed to
>
> none – not used because so set or nothing useable found
> value – the actual device to be used
>
> I still think it would be more readable if we had xsce_wan_enabled and
> xsce_lan_enabled (and get rid of xsce_gateway_enabled)
>
>
Think that would be about the same as using "none", just as a boolean.
Where should this entries be created and with what defaults?


> False – don’t use this interface even if you find it
> True – use whatever you find. can only use for wan if it is a gateway and
> for lan if it is not a gateway. auto is the default for both.
>
> and then assign none or an actual device to xsce_wan_iface and
> xsce_lan_iface based on enabled True/False and what is discovered.
>
>
Think you want to record the discovered interfaces as found and set the
network up based on the xsce_*an_enabled entries and record that result
also. I take it you're looking to do sanity checking on input from the GUI?
If so that is reasonable.



> > Named
> > is bound to all interfaces by default that is why there is the dispatcher
> >script that reloads named as interfaces come and go.
>
> so none of the scripts relies on our interface variables?
>

named, squid, dans don't use any interface names, while dhcpd(lan) and
wondershaper(wan) use appropriate interfaces names in their config files.


> >
> >
> >> I still find it confusing that xsce_lan_iface is sometimes state (none =
> >> none found) and sometimes a setting (none = don't use it). same for wan,
> >> and see first question. I have read through this logic 10 times and I
> still
> >> wonder if it catches every case.
> >>
> >
> >Just changed that because xsce_lan_iface can really be "" when detected as
> >such and discovered_lan is registered as soon as the shell code is run, it
> >is of no use to determine if there is no interface. see the last commit.
> >
> are you saying that shell: "/sbin/ip link show | grep -v -e DOWN -e lo -e
> tun -e {{ ansible_default_ipv4.alias }}| grep mtu | gawk -F: '{print $2}' |
> tr -d ' ' | head -n1 "
>
> can return “” and that we accept it if it does?
>

That would be valid if there is no detected LAN, we could set that to none
to reflect that fact. What would you like to have recorded in
/etc/sysconfig/xs_*an_device, "none" or ""? I'm leading towards recording
"none" and adjusting xs-gen-iptables to respect "none" in place of "".



>
> >
> >> what will be the result (value of xsce_wan_iface) for the following
> >>
> >> set_fact:
> >> xsce_wan_iface: "{{ xsce_wan_iface }}"
> >> when: 'xsce_wan_iface != "auto" or xsce_wan_iface != ""'
> >>
> >> when xsce_wan_iface = "none" - "none"?
> >> when xsce_wan_iface = "auto" - "auto"?
> >> when xsce_wan_iface = 192.168.0.25 - 192.168.0.25?
> >> when xsce_wan_iface = "" - ""?
> >> when xsce_wan_iface = "p3p5" - "p3p5"?
> >>
> >Only if you're messing with vars/*, trying to avoid the need for that and
> >still respect vars if present.
> >
> >
> >> it looks to me like xsce_wan_iface always ends up "{{ xsce_wan_iface }}"
> >>
> >Yes, when not set to auto or none in vars/* that is the point of being
> able
> >to use vars/.
>
> all I’m trying to say is this boils down to
>
> if (any condition) then A = A
>
> so if the condition is met A is set to A and if not A remains A. It
> doesn’t do anything.
>
> >
> >btw, after the next two declarations xsce_wan_iface can still be "auto" or
> >> anything besides none if ansible_default_ipv4.gateway is not defined.
> >>
> >Can you clarify that for me please. xsce_wan_iface is set to "none" when
> 'not
> >ansible_default_ipv4.gateway is defined'
> >
> sorry, I think I took an or for an and. so this really boils down to
>
> set_fact:
> xsce_wan_iface: "none"
> when: not ansible_default_ipv4.gateway is defined'
>
> (no need to set to none if it is already none.)


Think the conditionals can be slimmed down some.

tim-moody:
>
>> can they also be explicitly set to a value, like p2p7?

>I would prefer not to, lets think of that as expert mode, use at your
>own discretion.

agreed

>Think that would be about the same as using "none", just as a boolean.
>Where should this entries be created and with what defaults?

in default_vars:

xsce_wan_enabled: True
xsce_lan_enabled : True

in local_vars, change if desired

xsce_wan_iface and xsce_lan_iface at that point become simply working variables in computed vars, not user settable


>Think you want to record the discovered interfaces as found and set the
>network up based on the xsce_*an_enabled entries and record that result
>also. I take it you're looking to do sanity checking on input from the GUI?
>If so that is reasonable.

the gui won’t handle most of this.

the network has to be in place before the gui can even run.  so it assumes base has already been installed.

it is really just for changing the default configuration and I’d like to keep it simple, which is why I’m pushing for just two boolean variables that can be set by the user with a check box and the meaning of which can be explained relatively easily.

everything else derives from those settings.  of course the filters can be turned on and off from the gui, but not named or dhcpd, except implicitly by chosing a particular network configuration.

PR #222, opened 2015-02-22
RFC add xsce_lan_enabled xsce_wan_enabled

jvonau:
As a continuation of the conversation in #221  Is this the general way you envision how the concept should look? If this is looks good please close #221. I'm testing this over the next bit.

PR #224, opened 2015-02-23
dhcpd error

jvonau:
When the network is restarted the provided NM's dispatcher script restarts dhcpd when enabled in systemd but /etc/sysconfig/dhcpd might list stale data is one cause of the error noted in #222. Suppress this by restarting the network last there by refreshing the data and setting dis(en)abled before the script fires off. Remove the unconditional restart of dhcpd during installation just enable the service because the installation might be in single interface mode at that point, would be the other cause of the error noted.

### DIFF
--- a/roles/3-base-server/meta/main.yml
+++ b/roles/3-base-server/meta/main.yml
@@ -1,6 +1,6 @@
 dependencies:
-   - { role: network, tags: ['base','network'] }
    - { role: gateway, tags: ['base','gateway'] }
+   - { role: network, tags: ['base','network'] }
    - { role: httpd, tags: ['services','httpd','base'] }
    - { role: postgresql, tags: ['services','postgresql','base'] }
    - { role: authserver,  tags: ['services','authserver','base'] }   
--- a/roles/gateway/tasks/dhcpd.yml
+++ b/roles/gateway/tasks/dhcpd.yml
@@ -33,6 +33,7 @@
            state=stopped
   when: not dhcpd_enabled
 
+# service is started via NM dispatcher
 - name: Enable dhcpd service
   service: name=dhcpd 
            enabled=yes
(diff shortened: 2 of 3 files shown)


tim-moody:
we should have a comment that dhcpd is being started because it is in 
/etc/NetworkManager/dispatcher.d

jvonau:
comment added

jvonau:
the ordering change is in #228, will add NM note and adjust the service state during re-org

georgejhunt:
I'm thinking about our customer base.  Tim has been advising TK and I don't
know how many others to use master.

Maybe we should put in a tag before we change stuff at the networking
level, so if it takes a while to get it all sorted out, we can just say
"git clone --tag works --depth 1"

tim-moody:
Interesting, I was having the same thought.

how about a ‘stable’ tag, assuming we know where to put it?

georgejhunt:
If I downloaded something called "stable", I'd expect everything to work.

For master, until we go through the testing cycle, which I think is still a
little off, I'd like a name which communicates "likely to install, but
expect problems".

What that word is? Maybe "unstable".   --That is 5.1 is "stable", this new
tag is "unstable" or "new_features", and the head of master is "testing"

An aside, I pushed the first rpi2 image to unleashkids (1.8G). --very
rough, until the bridge stuff lands, I'm sort of reluctant to start
automating with ansible.

jvonau:
I'm about done #228-#230 the bridging was a last second thought but the
basics are there.

georgejhunt:
I just responded to TK who had a problem which may be related to using a
later/earlier version of ansible, The copy verb may have started needing
"/" at the end of the destination string (when it was actually a
directory), when it was not required before.

So this pushes me in the direction of a branch rather than a tag. I propose
that we  create a branch at current head called "new_features". Then we can
support him, and others, who need fixes.  We'd just need to be disciplined
about cherry-picking fixes into the "new_features" branch.

tim-moody:
another strategy would be to 

1) make sure master always installs, even if not everything works, which is pretty much the definition of your branch.  I thought this was already our policy.
2) wait until TK says he’s done installing and then take the commit hash from his xsce.ini and call it 5.5-pre-release or alpha (‘new features’ doesn’t tell me much)

PR #227, opened 2015-03-02
move network to be under gateway, add basic bridging 

jvonau:
move named, move all of network, add bridging. I can split them up if you prefer


jvonau:
Don't push this yet

georgejhunt:
226, and 227 are goiing to make my life a lot easier with rpi.

I'll go start on shrinking what I have that is working, and come back to
integrating when your stuff in in master.

If it's helpful:

- name: Count LAN ifaces
  shell: "/sbin/ip link show | grep -v  -e lo -e tun -e {{ xsce_wan_iface
}}| grep mtu | wc | gawk '{print $1}'"
  register: num_lan_interfaces
  ignore_errors: True
  changed_when: false

- name: If more than one LAN iface, what is name of second?
  shell: "/sbin/ip addr show | grep mtu | gawk -F: '{print $2}' | grep -v
 -e lo -e tun -e {{ xsce_wan_iface }} -e br0 | head -2 | tail -1"
  register: second_iface
  ignore_errors: True
  changed_when: False

jvonau:
closed in favor of #228 #229 #230 ran into an oops when I removed roles/network.main.yml in this one, should of left a placeholeder

tim-moody:
At some point can we get wan ip address override into this mix?

jvonau:
Right now we have a chicken egg situation. One needs to have internet
active to clone, so the static ip addressing would already be done to have
a gateway present. We're not messing with wan ifcfg files currently but
could. The questions become
1. should we alter or remove files to enable or disable the WAN? My vote is
alter
2. do we want to supply an ifcfg file if missing but gateway is active? My
vote is supply
3. write an ifcfg file for WAN when not detected using user static info.
Think as a starting point we'll need:
wan_ip:
wan_netmask:
wan_gateway:

have wan_ip: default to 'dhcp' for use in the new wan template

Thinking of using ifcfg-DEVICE where DEVICE is the real physical name.

tim-moody:
I think the use case is more like this:

1) I install xsce, a dhcp wan connection is fine.  accessing at schoolserver(.lan) is OK.
2) I want to deploy as an appliance with a fixed ip address, so now I want to change the wan ip address so that users can find it by ip or I can do special routing like put in a DMZ.  I do this through the console.

>From: Jerry Vonau 
>Sent: Tuesday, March 03, 2015 2:43 PM
>To: XSCE/xsce 
>Cc: Tim Moody 
>Subject: Re: [xsce] move network to be under gateway, add basic bridging (#227)
>
>Right now we have a chicken egg situation. One needs to have internet
>active to clone, so the static ip addressing would already be done to have
>a gateway present. We're not messing with wan ifcfg files currently but
>could. The questions become
>1. should we alter or remove files to enable or disable the WAN? My vote is
>alter

probably agree, but not sure of the issues

>2. do we want to supply an ifcfg file if missing but gateway is active? My
>vote is supply

I’m not convinced that the case of no ifcfg file exists.  It hasn’t yet for me.

>3. write an ifcfg file for WAN when not detected using user static info.
>Think as a starting point we'll need:
>wan_ip:
>wan_netmask:
>wan_gateway:

yes, and maybe consider a flag which says whether to use them or at least not a ‘none’/’auto’ value.

>
>have wan_ip: default to 'dhcp' for use in the new wan template
>
>Thinking of using ifcfg-DEVICE where DEVICE is the real physical name.

I think that’s what happens automatically, at least that’s what I always see.

>

georgejhunt:
My observation is that NetworkManager, and maybe also the old network,
apply dhclient to any newly discovered adapter.

 I've had situations where just deleting the ifcfg files, helped my offline
installs self configure properly (I think it was when the LAN adapter had a
different HWADDR because the image was applied to a new machine).

Now that we have a bridge getting all the setup details, and slaves
attached thereto, I think we should make ifcfg correspond with detected
hardware whenever ./runtag gateway is executed (whether it is delete, or
modify, seems almost an implementation issue). It's easier I think to
delete and then create.

tim-moody:
>I've had situations where just deleting the ifcfg files, helped my offline
>installs self configure properly (I think it was when the LAN adapter had a
>different HWADDR because the image was applied to a new machine).

even just removing the HWADDR line sometimes repairs it.

>Now that we have a bridge getting all the setup details, and slaves
>attached thereto, I think we should make ifcfg correspond with detected
>hardware whenever ./runtag gateway is executed (whether it is delete, or
>modify, seems almost an implementation issue). It's easier I think to
>delete and then create.

I take it you mean when the gateway role is run, regardless of by what means.

georgejhunt:
Yes, I suspect that once it's debugged, and complete, your console will be
the main only way that most users will interact.

Most of our installations will be headless I think. Does that mean the port
443 (80) need to be open on WAN, so that there might be two chances of
stimulating autoconfigure of networking from the GUI? No we surely don't
want 80 open. And we need to get xs-network-config (run gateway) into
scripts, so that a user can come in on port 22, or do what needs to be done
from the console.

tim-moody:
I have wondered about opening ssl on the wan.  if it’s an appliance we have said we will open everything.  if it’s a gateway, you can connect on the lan.  but I’m sure there will be lots of people who connect to the wrong nic and wonder why nothing came up.

I think it partly depends on what is upstream.  if there’s a firewalled router, then it’s not much different than the appliance case, but if it’s straight to the internet then it needs to be locked down.  there could also be school or gov’t policies that take precedence.

I guess a 3g dongle is the only case where the wan is open to the internet.

georgejhunt:
But the 3g dongle is the most likely in my opinion.  We've got 22 and 1194
(openvpn) open on WAN. It seems to me that we might address the "connected
mistakenly to WAN" by opening 443(will this give access to your console?)
when the source ip is a local ip address (could we let our dhcpd server
listen,and serve, on WAN adapter? Is that a security problem?)

georgejhunt:
dumb idea! can't have dhclient, and dhcpd server on the same adapter.

jvonau:
It seems you want to have port 443 open on WAN like ssh is. Would be best
if user can selectively disable either or both. Some might want to alter
the ssh port as commonly done. You can go direct to the console just use
https://<ip>/admin and login with 443 open. A 3g modem might be wide open
but depends on if the gateway role was applied.

tim-moody:
>
>From: Jerry Vonau 
>Sent: Wednesday, March 04, 2015 3:19 AM
>To: XSCE/xsce 
>Cc: Tim Moody 
>Subject: Re: [xsce] move network to be under gateway, add basic bridging (#227)
>
>It seems you want to have port 443 open on WAN like ssh is. Would be best
>if user can selectively disable either or both. Some might want to alter
>the ssh port as commonly done. 

as long as we prevent them from locking themselves out.

>You can go direct to the console just use
>https://<ip>/admin and login with 443 open. A 3g modem might be wide open
>but depends on if the gateway role was applied.

if you know the ip.  but good point, the docs could say https://schoolserver/admin rather than http

PR #228, opened 2015-03-03
move named to gateway role

jvonau:
part one of the moving network to gateway. tested runs fine


jvonau:
Part 2 in new PR

jvonau:

PR #229, opened 2015-03-03
Move network to gateway part2 of #228

jvonau:
build on top of PR #228, tested fine.


jvonau:

PR #230, opened 2015-03-03
Bridge2

jvonau:
Built on top of #229 work to do is lan detection for more that one interface. stopping here for now. runtags gateway does the right thing...


jvonau:

PR #232, opened 2015-03-05
RFC Tagging

jvonau:
This allows skipping of prep.yml and prelim.yml after the first run of ansible, computed_* facts are still done under the platform tag.

### DIFF
--- a/roles/1-prep/tasks/main.yml
+++ b/roles/1-prep/tasks/main.yml
@@ -1,9 +1,18 @@
 - include: prep.yml
+  tags:
+    - prep
+
 - include: computed_vars.yml
+  tags:
+    - platform
+
 - include: computed_network.yml
+  tags:
+    - platform
            
 # Put conditional includes for platforms here
 - include: xo.yml
   when: xo_model != "none"
   tags:
+    - platform
     - xo
--- a/roles/2-common/tasks/main.yml
+++ b/roles/2-common/tasks/main.yml
@@ -2,13 +2,25 @@
   when: ansible_selinux["status"] != "disabled"
   tags:
     - download
+    - common
     
 - include: prelim.yml
+  tags:
+    - prep
 
 - include: yum.yml 
   tags:
+    - common
     - download
 
 - include: chrony.yml
+  tags:
+    - common
+
 - include: avahi.yml
+  tags:
+    - common
+
 - include: sysctl.yml
+  tags:
+    - common
--- a/runtags
+++ b/runtags
@@ -22,7 +22,7 @@ found="N"
 
 for tag in $tags
 do
-  if [ "$tag" == "prep" ]
+  if [ "$tag" == "platform" ]
   then
     found="Y"
   fi
@@ -35,7 +35,7 @@ taglist=$1
 
 if [ "$found" == "N" ]
 then
-  taglist="prep,"$taglist
+  taglist="platform,"$taglist
 fi
 
 # script to run a specific tag in the current playbook
--- a/xsce-base.yml
+++ b/xsce-base.yml
@@ -8,5 +8,5 @@
 
   roles:
       - { role: 1-prep, tags: ['prep','platform','base'] }
-      - { role: 2-common, tags: ['common','base'] }
+      - { role: 2-common, tags: ['prep','common','base'] }
       - { role: 3-base-server, tags: ['base'] }
--- a/xsce-from-console.yml
+++ b/xsce-from-console.yml
@@ -8,7 +8,7 @@
   - /etc/xsce/config_vars.yml
 
   roles:
-      - { role: 1-prep, tags: ['prep','platform','base'] }
+      - { role: 1-prep, tags: ['platform','base'] }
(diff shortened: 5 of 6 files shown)


tim-moody:
My experience is that tags are not inherited, making them almost useless.  So if I tag 

- {role: xxx, tags: yyy} and runtags yyy, only things in xxx that are not explicitly tagged something else (or are tagged yyy) will run.  (If this is not the case, please let me know, but it’s what my experiments showed.)

so using tags is like a programming language in which flags can be set on each line of execution and those lines are executed depending the flags.  I worked with such a language, RPG II, in the 80s and it was a real mess.

I don’t use and don’t plan to use tags in the console other than I have a utility for expert users which is a visual equivalent of runtags, but I don’t expect it to be the main way of doing things and there will be warnings about it.

From: Jerry Vonau 
Sent: Thursday, March 05, 2015 7:00 AM
To: XSCE/xsce 
Subject: [xsce] RFC Tagging (#232)

This allows skipping of prep.yml and prelim.yml after the first run of ansible, computed_* facts are still done under the platform tag.


  https://github.com/XSCE/xsce/pull/232

Commit Summary
  a.. add tags to 1-prep 
  b.. add tags to 2-common 
  c.. run prep tag in 2-common only during install 
  d.. run prep tag only during install, use platform to trigger computed_* 
File Changes
  a.. M roles/1-prep/tasks/main.yml (9) 
  b.. M roles/2-common/tasks/main.yml (12) 
  c.. M runtags (4) 
  d.. M xsce-base.yml (2) 
  e.. M xsce-from-console.yml (2) 
  f.. M xsce.yml (2) 
Patch Links:
  a.. https://github.com/XSCE/xsce/pull/232.patch 
  b.. https://github.com/XSCE/xsce/pull/232.diff

jvonau:
On Thu, Mar 5, 2015 at 7:56 AM, Tim Moody <notifications@github.com> wrote:

> My experience is that tags are not inherited, making them almost useless.
> So if I tag
>
> - {role: xxx, tags: yyy} and runtags yyy, only things in xxx that are not
> explicitly tagged something else (or are tagged yyy) will run. (If this is
> not the case, please let me know, but it’s what my experiments showed.)
>
>
That is what I'm observing, playbooks without tags in them are executed in
full or just the tagged yyy parts when tags are used. Seems like the
tagging is useful when you have a specific code path needed like how the
XOs are handled.


> so using tags is like a programming language in which flags can be set on
> each line of execution and those lines are executed depending the flags. I
> worked with such a language, RPG II, in the 80s and it was a real mess.
>
>
Seems like a shame not to use tags for some operations that only need to be
done once like creating directories. Makes for a shorter operation and a
cleaner log when you're not repeating operations that are already
performed. Just an idea.


> I don’t use and don’t plan to use tags in the console other than I have a
> utility for expert users which is a visual equivalent of runtags, but I
> don’t expect it to be the main way of doing things and there will be
> warnings about it.
>
>
You are using tags indirectly in the console by the use of
xsce-from-console.yml. That is what I see running when I press "install
configured server options". Are you planning on changing that?

tim-moody:
I think tags at the level of role could be useful if there are no tags below.  I don't think we need tags on things like directory creation because ansible does checks for existence. We need to take advantage of the declarative nature of ansible that says what should be true, not what should be done. The biggest inefficiency I see is the step in common that says make sure all rpms are up to date. That has to be checked on every run, but I assume that once base is done it doesn't need to run again unless you change networking.Sent from my BlackBerry® PlayBook™www.blackberry.comFrom: "Jerry Vonau" <notifications@github.com>To: "XSCE/xsce" <xsce@noreply.github.com>CC: "Tim Moody" <tim@timmoody.com>Sent: 5 March, 2015 11:06 AMSubject: Re: [xsce] RFC Tagging (#232)On Thu, Mar 5, 2015 at 7:56 AM, Tim Moody <notifications@github.com> wrote:

> My experience is that tags are not inherited, making them almost useless.
> So if I tag
>
> - {role: xxx, tags: yyy} and runtags yyy, only things in xxx that are not
> explicitly tagged something else (or are tagged yyy) will run. (If this is
> not the case, please let me know, but it’s what my experiments showed.)
>
>
That is what I'm observing, playbooks without tags in them are executed in
full or just the tagged yyy parts when tags are used. Seems like the
tagging is useful when you have a specific code path needed like how the
XOs are handled.


> so using tags is like a programming language in which flags can be set on
> each line of execution and those lines are executed depending the flags. I
> worked with such a language, RPG II, in the 80s and it was a real mess.
>
>
Seems like a shame not to use tags for some operations that only need to be
done once like creating directories. Makes for a shorter operation and a
cleaner log when you're not repeating operations that are already
performed. Just an idea.


> I don’t use and don’t plan to use tags in the console other than I have a
> utility for expert users which is a visual equivalent of runtags, but I
> don’t expect it to be the main way of doing things and there will be
> warnings about it.
>
>
You are using tags indirectly in the console by the use of
xsce-from-console.yml. That is what I see running when I press "install
configured server options". Are you planning on changing that?

jvonau:
Closing in favor of #234

PR #233, opened 2015-03-05
improvements for iptables & avahi

jvonau:
1. allow user to use other than stock ssh port(actual port change in sshd is not covered)
2. allow access to https for console
3. allow access for avahi on all interfaces
4. use preexisting flag to en/disable forwarding
5. advertise ssh and https services with avahi
6. lock down squid when dansguardian is active
7. optionally intercept DNS bound for the web.

Note: tested xs-gen-iptables routine, other template matches but not tested.


jvonau:

PR #245, opened 2015-03-22
Mar22 take2

jvonau:
Replacing PR #244


georgejhunt:
Does this mean that kiwix is not available for i686?

jvonau:
There is no i686 listed to download or on the website.

tim-moody:
The changes to kalite will break the install.

The principal is that we always install and then either enable or disable services.  kalite needs to be running in order for the install to complete.  that’s why it was started in two places.

jvonau:
How does this break the install when you always have to run install.yml
then always run enable.yml? What will fail is if you don't install kalite
by default then try to disable the non-existent service in enable.yml

jvonau:
I'm closing this one as the last 2 commits were unintended going to redo

tim-moody:
sorry, think I confused this with owncloud where the service actually has to be running during the configuration step of the install.

PR #248, opened 2015-03-25
Mar25 fixes

jvonau:
Replaces PR #245

### DIFF
--- a/roles/1-prep/tasks/computed_vars.yml
+++ b/roles/1-prep/tasks/computed_vars.yml
@@ -91,3 +91,14 @@
     mysql_service: mysqld
   when: ansible_distribution_release == "based on Fedora 18"
 
+- name: Set kiwix source file name x86_64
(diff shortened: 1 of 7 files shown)


tim-moody:
I'm of two minds on error trapping like is done for kiwix.  Not having errors which cause the thing to halt is always good, but here you will get no notice if kiwix download is not found and will be left wondering what happened.  Can we have a debug message?

jvonau:
The behavior didn't change, the playbook doesn't run if kiwix_src_file is
undefined. Didn't think a error was needed if run on i686, you'll see
"skipping".

tim-moody:
if it’s undefined you’re on an unsupported platform, so I think it should crash.  There are hundreds of skipping messages.

jvonau:
meet halfway with a warning, no fail

jvonau:
Replaced by PR #252

PR #257, opened 2015-03-31
egrep of the mysql task does not fine enablement

georgejhunt:
on my NUC, and after a reboot, mysql was not running. But systemctl start mariadb started it successfully


georgejhunt:
I saw the state=started, but now I realize that it needs to be started in order to configure it. But then separately, the decision needs to be made to enable it.  I defer to Tim, since I see he has a plan

PR #258, opened 2015-04-01
RFC: postgresql handling

jvonau:
Tim: is this what you envisioned? Maybe move postgresql and authserver to 5-xo-services?


jvonau:
Now part of #264

PR #256, opened 2015-03-30
Nm fix

georgejhunt:
Found these changes necessary for install of master on XO4

PR #259, opened 2015-04-01
enlarge /tmp file system

jvonau:
Proposed replacement of #256


tim-moody:
does this require further testing or is it ready to merge?

georgejhunt:
I have not actually merged it, and tested the ansible playbook,(but I did
change fstab by hand, and verify that it was necessary, and sufficient) but
my vote is to go with it now

PR #261, opened 2015-04-01
generate and announce the uuid for this install

georgejhunt:
As I've been looking at further automating the testing of the server, I've been wanting to merge the results from the tests done be an XO client and the tests tests that the server does on itself. Having a UUID to link data from both seems like a good idea.  I'm looking at couchdb on unleashkids.org as one option.

### DIFF
--- a/roles/1-prep/tasks/computed_vars.yml
+++ b/roles/1-prep/tasks/computed_vars.yml
@@ -50,6 +50,8 @@
       value: '{{ ansible_local["local_facts"]["xsce_branch"] }}'
     - option: 'xsce_commit'
       value: '{{ ansible_local["local_facts"]["xsce_commit"] }}'
+    - option: 'xsce_uuid'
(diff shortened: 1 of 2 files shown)


jvonau:
I like the idea but your missing a :  in  local_facts.fact.j2

georgejhunt:
thanks, fixed

tim-moody:
why not tie to the hash of the commit of the version installed found in xsce.ini?  basically you are saying here are the results for this version of the server.  the presence or absense of content could muddy the water, but that's a topic we need to discuss in terms of test data.

tim-moody:
I guess if you had two install runs with different config settings then the commit wouldn't change, but the install would.

If the db is single user I like sqlite3.  It's what the cmsrv uses.

jvonau:
Think you might want to use UUID as a deployment identifier,  commit for the version,  date for general info

PR #262, opened 2015-04-01
RFC proposed changes

jvonau:
For Friday's discussion, there are a few ideas that are in the PR. 
<<< are comments, the rest of the service are populated after running the console.   
The xsce.ini  file after install-console looks like:

[location]          <<< info in this section will never change.
xsce_base = /opt/schoolserver
xsce_dir = /opt/schoolserver/xsce

[version]         <<< info in this section will never change.
xsce_branch = install
xsce_commit = d43b79526d8e70930039c2985fe1c3e5c439262a
install_date = 2015-04-01T18:31:12Z    <<< This would be the time of the first run
install_xo = none

[runtime]           <<< The rest of the sections updated during the run.
runtime_branch = install
runtime_commit = d43b79526d8e70930039c2985fe1c3e5c439262a
runtime_date = 2015-04-01T22:14:29Z     <<< This would be the time of the last run.
runtime_php = /usr/lib64/php
runtime_preload = True      <<< used to toggle use_cache, no_network, and xsce_prepped

[network]
wan_enabled = True
lan_enabled = True
detected_gateway = p2p1
detected_lan = p7p1
user_gateway = auto
user_lan = auto
computed_gateway = p2p1
computed_lan = p7p1
xsce_network_mode = Gateway

[xs-authserver]
name = XS-authserver
description = xs-authserver implements a seamless web authentication service using XO laptop registration capabilities.  It is heavily inspired by the Moodle OLPC-XS authentication plugin
port = 5000
path = /


tim-moody:
xsce.ini is backed up when modified, so earlier values are still around.

 

I'm not clear on why you want runtime to be different from version.  version should change if you do a pull.  what use will be made of the original versions?

 

how will you determine the first run?  what if there is an xsce.ini already there from a usb offline install?

 

runtime_php is computed based on os.  when will you use the stored value and when the computed value?

 

what does install_xo mean?

 

no network is a fact not a setting.

 

my other comment is that we are becoming more and more procedural

jvonau:
On Wed, Apr 1, 2015 at 7:54 PM, Tim Moody <notifications@github.com> wrote:

> xsce.ini is backed up when modified, so earlier values are still around.
>
>
>
Successful runs will be of no interest to us, we will only hear about
failures and that will be the last run. Keeps the filesystem neater.


>
> I'm not clear on why you want runtime to be different from version.
> version should change if you do a pull. what use will be made of the
> original versions?
>
>
Kind of nice to see how long this machine has been deployed without further
digging.


>
>
> how will you determine the first run?


Both xsce.yml for runabsible & runtags and xsce-base.yml for
install-console start at 0-once.
Either will trigger the creation of the file. Note that
xsce-from-console.yml skips steps 0,2,3 speeding up the run.


> what if there is an xsce.ini already there from a usb offline install?
>
>
>
Then new info is add and values are changed to reflect the current run.
When creating images I use runtags=download,download2 to pre-populate the
image adding /.preload as a marker.



> runtime_php is computed based on os. when will you use the stored value
> and when the computed value?
>
>
The underlying fact or its use didn't change this is more for just
recording the detected value, could be in the install section or dropped,
that is why this is an RFC.


>
>
> what does install_xo mean?
>
>
Recording if the install in on an XO.


>
>
> no network is a fact not a setting.
>
>
>
lol... we're setting a fact no?


>
> my other comment is that we are becoming more and more procedural
>
>
>
In a coding way or in the way we interact with each other?

tim-moody:
can this be closed?

jvonau:
incorporated into #272

PR #265, opened 2015-04-04
Reordering

jvonau:
Cleaned up version of #262, replacing #264


tim-moody:
we agreed to use meta in the role that requires mysql (and postgre), but you have taken it out

jvonau:
Yes, we are always installing both of those services, but leaving them disabled. We then set *enabled in computed_vars based on dependencies when run, this work great with xsce-from-console.yml and xsce.yml, both services' playbooks are only run once. What would be an issue is running 'runtags <dependent service>' without having either xsce-from-console.yml or xsce.yml being run first to set the services as enabled, the workaround would be to include the DB in the call to runtags. I feel the speed up with this change is justified.

jvonau:
replaced by #266

PR #266, opened 2015-04-04
This gets us to a livecd, on the way to imaging

jvonau:
Replacing #265


jvonau:
incorporated into #272

PR #263, opened 2015-04-03
RFC -- these are the Raspberry Pi 2 networking changes

georgejhunt:
Assignment of adapters works differently now:
  1. look for wifi, and gateway
  2. count the number of adapters that are not wifi and not gateway
  3.  if wifi plus this number > 1, create a bridge, and connect slaves to bridge
If hostapd/wifi/bridge is involved, enable network.service as well as NM.
Add and subtract lines in /etc/rc.d/rc.local to attach adapters to bridge.
Add a hw-detect yml file

PR #267, opened 2015-04-04
RFC - FYI tracking jv's installer4 branch

jvonau:
This is based off of #266 and is not meant to merged. This is more of a heads up, tracking work in progress.

PR #268, opened 2015-04-06
RFC for testing only rpi2 logic, bridging code, full wifi-AP support

jvonau:
This workup is based on Georges' PR #263. The networking bridging code works for me on a F18 trimslice so should work straight away. This would form the bases of supporting WiFi-AP with or without bridging. The only issue I had is with the detection of the wifi device with this code:

shell: "cat /proc/net/wireless | grep -v -e Inter -e face | gawk -F: '{print $1}' | head -n1"

I think this is a kernel thing with F21 vs F18-F20 which returns blank, I'd like to confirm this with other. I don't have a F21 box handy so I would like some help. This is what cat proc/net/wireless returns:

Inter-| sta-|   Quality        |   Discarded packets               | Missed | WE
 face | tus | link level noise |  nwid  crypt   frag  retry   misc | beacon | 22

The good news is that if you set  {discovered_wireless_iface} manually in vars everything works on F18-F20. George I would really appreciate it if you could give this a spin, and tell me if the changes work for you.


jvonau:
Please don't merge this after testing, I'd like to see PR #266 get in first, then I can rework this one for the altered layout, other wise every thing in PR #267 might need redoing.

georgejhunt:
I find it frustrating to have spent essentially 2 weeks of testing and
refining, and then to need to start over testing your new branch, which
admittedly based upon some of my early work.

My tests are documented at
https://docs.google.com/spreadsheets/d/18Y07kuAiHyMGJNQLydgo7GbuHShu_F5BQWA3egt_1PU/edit#gid=0
since I can't remember what I've done. I suspect that I'll need to
cherrypick fixes into your branch.  But I have not been disciplined about
one fix per commit.

I'm a little discouraged. I'll start testing on the rpi2 tomorrow.

jvonau:
TASK: [network | Enable the Access Point Hostapd program] ********************* 
[91956.485080] device wlan0 entered promiscuous mode
changed: [127.0.0.1]

TASK: [network | Disable the Access Point Hostapd program] ******************** 
skipping: [127.0.0.1]

TASK: [network | Enable network.service for non-bridged Wifi] ***************** 
skipping: [127.0.0.1]

TASK: [network | Disable network.service for bridged Wifi] ******************** 
[91957.454687] IPv6: ADDRCONF(NETDEV_UP): wlan0: link is not ready
[91957.672929] IPv6: ADDRCONF(NETDEV_CHANGE): wlan0: link becomes ready
[91957.679842] br0: port 2(wlan0) entered forwarding state
[91957.685332] br0: port 2(wlan0) entered forwarding state
[91957.928122] device wlan0 left promiscuous mode
[91957.932858] br0: port 2(wlan0) entered disabled state
ok: [127.0.0.1]

The last few commits were trying to deal with an intermittent (20%) failure adding the wifi device to the bridge.

georgejhunt:
Did 4 tests on my rpi2 this morning. gateway and lan controller with 1 - 3 adapters in bridge worked fine.
But appliance balked at getting rid of the bridge, and came up with lan controller config.  Some details are at https://docs.google.com/spreadsheets/d/18Y07kuAiHyMGJNQLydgo7GbuHShu_F5BQWA3egt_1PU/edit#gid=0
My "mypi branch has clauses that deal with this in 1-prep/tasks/computed_network.yml . I suggest you look at (and perhaps cherry pick) some of b2c10, 5f3cdf, -- the rebooting of NM in 5f3c was in an attempt to find a gateway if we could. I don't think that worked. But I never pulled it back out.

jvonau:
Having a bridge with no ip address and no slaves is not really a problem,
br0 would be like any other interface just sitting there. There might be
the possibility that the bridge might grab an ip address from dhcp on the
server, but that would be about it. The same thing can occur with an unused
but cable attached wired interface.

The bridge could be viewed as an always attached interface, needing
xsce_lan_enabled to be set to False to override the detection to enable
appliance mode. The main issue in appliance mode would be iptables loading
because of the presence of the bridge, think it could be just easier to
clear the iptables rather than messing around deleting/changing ifcfg
files. Could do something like {if gateway and no slaves then clear
firewall}, but if there is a slave present then xsce_lan_enable should be
set to False by the user.

I pushed a couple of untested commits (proposed gateway re-detection) for
your comments. I'll test tomorrow night, I'm little busy this week, picked
up a short term contract.

jvonau:
Shoot, forgot I'm always forcing bridging for the rpi2 by setting
use_bridging: True. Commit 0ef8d85d would reverse that but you would need
to add use_bridging: True to local_vars to enable bridging until that
option is available in the console.

jvonau:
incorporated into #272

PR #275, opened 2015-04-15
Apr13 fixes2

jvonau:


georgejhunt:
Jerry,
I got a failure to start network.service. Log said something about not being able to join an adapter to the bridge because it already belonged to a bridge.  I changed ifcg back to NM_managed =no, and restarted both NM and network successfully.

PR #276, opened 2015-04-15
Dbstartup

georgejhunt:
Computed enabling of databases only works if those db roles run when the dependent role run


tim-moody:
rachel already there

georgejhunt:
Had forgotten that rachel used mysql. I updated PR

PR #279, opened 2015-04-16
after install-console, cannot acces from remote browser, firewalld is ru...

georgejhunt:
Starting from fresh netinst x86_64, NUC, running asnible ./install_console, cannot access console from remove browser because firewalld is still running. Network discovery had decided appliance (I forgot to plug in the USB dongle).

### DIFF
--- a/roles/network/tasks/iptables.yml
+++ b/roles/network/tasks/iptables.yml
@@ -1,6 +1,7 @@
 - name: Disable firewalld service
   service: name=firewalld
            enabled=no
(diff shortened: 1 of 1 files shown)


jvonau:
Yea, I can see that, firewalld would be active during the install. FWIW,
the livecd installer would of disabled that during the initial install
because it's kickstart file said to. A reboot after install-console would
of allowed the change to take effect. Merging

jvonau:
merged #279

PR #282, opened 2015-04-17
give kalite-serve longer to start up

georgejhunt:
Had a failure on the 2830 NUC while kalite-serve was starting up. Logs not helpful. Started ok from the command line, but took about 30 seconds.

### DIFF
--- a/roles/kalite/tasks/enable.yml
+++ b/roles/kalite/tasks/enable.yml
@@ -8,6 +8,8 @@
   service: name=kalite-serve 
            enabled=yes
            state=started           
+  async: 60
+  poll: 5
   when: kalite_enabled  
 
 - name: Disable kalite cron server


tim-moody:
TK has said a number of times that kalite does not automatically start on boot.
I gave it more time in the service definition (90 seconds I think.), but that didn't solve the problem. Difficult because it is intermittent.  I wonder if there is a dependency that has to start first.

jvonau:
This looks fine for the initial install, go with it. Tim: The unit file has: TimeoutStartSec=120, but the pid file lives on the disk, network target can be complete before the filesystem is mounted, syslog is a dummy file. Think we might want to use local-fs.target or something a little later in the systemd boot sequence.

tim-moody:
I guess 90 was the default and I upped it to 120.  I think this is one of the unit files I cobbled together so feel free to fix it.

 

Probably I started with idmgr.

PR #285, opened 2015-04-20
disable iptables in appliance, enable it in gateway

georgejhunt:
My earlier attempt to disable firewalld was not successful. After clean install of i686 on i3 NUC, and running ./install-console, firewall was enabled, and iptables was disabled --- not ready for a reboot.

I added in the flag for iptables_enabled so that we could discuss it. I think it's needed, but I'm wrong lots of the time

### DIFF
--- a/roles/1-prep/tasks/computed_network.yml
+++ b/roles/1-prep/tasks/computed_network.yml
@@ -120,7 +120,6 @@
     named_enabled: False
     dhcpd_enabled: False
     wondershaper_enabled: False
-    iptables_enabled: False
     xsce_network_mode: "Appliance"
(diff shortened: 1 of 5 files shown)


jvonau:
I was thinking of leaving iptables always enabled to protect the databases in PR #284

tim-moody:
as long as the iptables definitions swap between network roles, I like leaving it enabled

I thought firewalld was turned off quite awhile ago in what is now network restart.yml.

georgejhunt:
Ok, I modified PR285 to just address the ineffective disabling of
firewalld, and the missing enabling of iptables.

PR #286, opened 2015-04-20
last minute fixes, tested on vm

tim-moody:
cmdsrv - get storage: ignore extended partitions, when > 4 partitions
console - fix race condition in async ajax calls


tim-moody:
Please note that cmdsrv error handling for non-existent commands is turned off for debugging.

PR #291, opened 2015-04-20
since services can be disabled at GUI, add enabled to ini values

georgejhunt:
I think this completes the transform of ini, so that it specifically mentions the enabled state of each service  -- So that presence in ini file signifies "installed" and enabled is separately visible


tim-moody:
rachel?

jvonau:
rachel is handled in rachel_enabled.yml

georgejhunt:
I'll add rachel

PR #300, opened 2015-04-23
Machine

georgejhunt:
I'm trying to automate the record keeping for my tests -- arch is important variable


georgejhunt:
ansible has this info, and more, and would be a better approach

PR #301, opened 2015-04-23
maybe openssl will fix the stall during pear install on xos

georgejhunt:
The stall on setting the pear directory on XO4 and XO1.5 was corrected. Earlier I had just done a "yum update" and caused this error to disappear. It's not clear to me whether it is better to update all of fc18 on XO's, or just the minimal set. We can start with the latter.  I don't know why James Cameron did not update everything with his new spin.

### DIFF
--- a/roles/2-common/tasks/yum.yml
+++ b/roles/2-common/tasks/yum.yml
@@ -27,6 +27,7 @@
    - hostapd
    - wpa_supplicant
    - wget
+   - openssl   #FC 18 does not supply, but pear requires
   tags:
     - download
 


quozl:
@georgejhunt, Because we use frozen repositories and have to test every change, and had no buy-in from community to do that testing.  If you think a change is needed, make sure a bug is raised at http://dev.laptop.org/ and CC me, thanks.  Refer to this PR.

http://wiki.laptop.org/go/Frozen_repositories
http://mock.laptop.org/cgit/
openssl-libs-1.0.1e-4 is in our sample builds.
openssl-libs-1.0.1e-37 is in koji for fedora 18.

PR #302, opened 2015-04-23
add in a 5 minute delay while kalite installs the database

georgejhunt:
Tested on i3NUC, I put in a 300 second delay, it reported state every 10 seconds, and used 80 seconds before completion.

Tim mentioned in an email that he thought Jerry had put in a delay. But I don't find it.

### DIFF
--- a/roles/kalite/tasks/install.yml
+++ b/roles/kalite/tasks/install.yml
@@ -43,6 +43,8 @@
             
 - name: Run the install
   command: "/tmp/kalite_install.exp"
+  async: 300
+  poll: 10
 
 - name: Create kalite service(s) and support scripts
   template: backup=yes


jvonau:
Push away!! I'm looking at systemd failures that could occur where we write
a pid file on the filesystem before having the filesystem mounted. work in
progress

tim-moody:
I'm still skeptical about using async instead of delay.  My reading of that is that it continues on in which case it will come to the restart.Sent from my BlackBerry® PlayBook™www.blackberry.comFrom: "Jerry Vonau" <notifications@github.com>To: "XSCE/xsce" <xsce@noreply.github.com>Sent: 23 April, 2015 9:23 PMSubject: Re: [xsce] add in a 5 minute delay while kalite installs the database (#302)Push away!! I'm looking at systemd failures that could occur where we write
a pid file on the filesystem before having the filesystem mounted. work in
progress

PR #303, opened 2015-04-24
Systemd unit files & iptables

jvonau:


georgejhunt:
So the question for me is: should we put "tested"branch before or after these hit. Do you think systemd timing is causing unreliable startup?

PR #308, opened 2015-04-28
Rpi2

georgejhunt:
These changes found to be necessary to our "stable" branch on rpi2. The wireless detection works on fc18-fc21, and on centos.


georgejhunt:
I disabled the network service change in this PR, hoping that rescanning the ifcfg files will work in the fc21 rpi2.

PR #314, opened 2015-05-01
Network

jvonau:
Change in NM behavior for F21 needing a reread of ifcfg files, with forced bridge removal.


georgejhunt:
Adam is wanting me to provide some way for one of his deployments to use wifi as gateway. So I think your test needs to include the possibility that wifi might be WAN, and therefore not part of a bridge

georgejhunt:
No, I read through the code again, and I don't think wireless _lan_iface will be set if I declare in local_vars that wlan0 is user_wan_iface.

PR #317, opened 2015-05-19
Sysctlfix

georgejhunt:
I'll pull this into master and test it in a single test. Did it by hand on an install, and it fixed runansible"


georgejhunt:
This was to help centos 6.8 load XSCE, but 6.8 does not have systemd, and that seems to add lots of complexity for little benefit

PR #320, opened 2015-06-12
Console jun4

tim-moody:
not sure if zim 3rd step works, but the rest has been tested


georgejhunt:
Do any of the changes you pulled into master require starting over? Or can
I just rerun the playbook? (one time in the past the old database structure
was an issue)

tim-moody:
no change to db

tim-moody:
most of it is non-console in-spite of name

PR #321, opened 2015-06-14
Add dokuwiki to install from the latest stable version

m-anish:
Notes:
* Is enabled to install by default. 
* There is no corresponding entry in default_vars.yml but the variables dokuwiki_installed and dokuwiki_enabled exist.
* Tested on centos x64 VM
* The whole /opt/dokuwiki directory is given write permission by apache, which is a bit unsecure. It is a PITA to give specific directories permissions, but will get around to it. 
* Suitable for merging in a testing branch, but not release. However, it does NOT break anything and is stable.
* Would like feedback on whether directories which will be edited due to user actions should lie under some specific subdirectory like /library.

Signed-off-by: Anish Mangal <anishmg@umich.edu>


tim-moody:
Excellent.  Glad to see this.

m-anish:
Great. Thanks for the feedback. Closing this PR and will open up a new one with the changes you suggest.

m-anish:
While I'm working on this, how should is best facilitate 'enabling' or 'disabling' of this. In debian, in apache, there is sites_enabled and sites_available. In EL7/Fedora, I guess there is only once place for conf files. Also, enabling dokuwiki is really just the 'Alias' directive, unlike other services like kiwix.

tim-moody:
elsewhere I just create or delete the .conf file in /etc/httpd/conf.d to enable or disable.

 

for example for owncloud

 

- name: Remove httpd conf file if we are disabled

  file: path=/etc/httpd/conf.d/owncloud.conf

        state=absent

  when: not owncloud_enabled.yml

 

the template is copied in owncloud_enabled.yml which is only included when: owncloud_enabled

tim-moody:
of course any link to the alias will just give a 404 if it is disabled, so it really needs to be integrated with the home page

PR #322, opened 2015-06-14
Add Opensearchserver to XSCE

m-anish:
* Retrieves the rpm package and stores it in downloads
* Installs the rpm. Enables the service
* Configures the server to run at port 9092 instead of 9090 (on which cockpit runs in fedora)
* Tested on CentOS 7 x64 VM
* Enabled by default
* Enable/Disable variables created but not really used - also not added to default_vars.yml


m-anish:
Ugh. Looks like this PR will also install dokuwiki :-(

PR #323, opened 2015-06-14
Add Opensearchserver to XSCE

m-anish:
* Retrieves the rpm package and stores it in downloads
* Installs the rpm. Enables the service
* Configures the server to run at port 9092 instead of 9090 (on which cockpit runs in fedora)
* Tested on CentOS 7 x64 VM
* Enabled by default
* Enable/Disable variables created but not really used - also not added to default_vars.yml


m-anish:
I just realized that this won't work on ARM, but only x86. So use for testing until the playbook is updated to check for arch.

georgejhunt:
I think we decided this morning to leave this PR open, so that we can play with it, and pull it into a testing environment, and perhaps look for some similar function which will work on ARM as well for the long term

m-anish:
Sounds good.

davelab6:
This is pretty old, should it be closed?

PR #326, opened 2015-06-18
Add dokuwiki to XSCE 

m-anish:
### Notes
* Is enabled to install by default. Gets the latest tarball from the official dokuwiki download location.
* There is no corresponding install/enabled variable entry in default_vars.yml but the variables dokuwiki_installed and dokuwiki_enabled exist AND are used by the playbook.
* Tested on centos x64 VM

#### Changes over previous PR 
* dokuwiki_url is now used as a variable
* dokuwiki now installs in /library instead of /opt
* This _should_ be mergeable :-)


georgejhunt:
I'll make the changes to point to unleash kids, and change enabled: False once it is in master.  I have not learned how to modify someone else's PR

m-anish:
Thanks!

PR #331, opened 2015-06-21
Cons 5.5-3

tim-moody:
Improve visual feedback on buttons
     Add reporting of error condition from server (not the ajax error already handled)
     Add button greying/ungreying to command calls and callbacks
     Add alerts
     Add more command info to job status return from server
     
Known Problems
     Refresh zim lib - ? does this do anything
     Restart kiwix – fails if no content
     Cancel Jobs  - can sometimes fail to update all wip zims in Get Zims from Kiwix screen
     Refresh Jobs - refreshed list is feedback, but doesn't always change except for last refreshed time.
     An alert would be a nuisance.
    
Fix a couple of bugs in cmdsrv


tim-moody:
Install and Code tested on 64 bit VM on Fedora 21.

PR #349, opened 2015-07-14
Awstats

georgejhunt:
I want to pull this into a test install, to check for typo's before I merge it into master

### DIFF
--- a/roles/awstats/defaults/main.yml
+++ b/roles/awstats/defaults/main.yml
@@ -1,0 +1,3 @@
+awstats_install: True
+awstats_enabled: False
+awstats_data_dir: /library/awstats/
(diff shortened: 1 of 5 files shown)


jvonau:
I would just change the squid-xs.conf for the logging changes, or we risk
loosing the change if squid's playbook is rerun.

tim-moody:
+#
+DirData="/var/lib/awstats" in the config looks wrong

where is the url to access this?

georgejhunt:
As you point out, changing squid-xs.conf will just get wiped out if the
playbook is run again. I don't think we have a clean way to deal with one
service modifying another. I'll just comment it out for now.

georgejhunt:
At line 203 of ./templates/apache-awstats.conf it is set to what I wanted.
Where did you see it set to /var/lib/awstats? I must have missed it.

tim-moody:
OK.  I must have been looking at one of the earlier commits before you set that.

PR #350, opened 2015-07-14
Teamviewer

georgejhunt:
It would be nice to have this in master for Adam, to build up his NUC.

Again, I'll test it before I merge it


jvonau:
This needs a conditional based on arch as the downloaded rpm is i386. Can
one download an x86_64 or arm version of the rpm?

PR #352, opened 2015-07-17
don't modify squid from awstats

georgejhunt:
I'll modify the squid config file template, when I focus on that.


tim-moody:
nope.  the syntax is in ansible and faults even if the step is not executed because it happens at parsing.

 

[root@xsce xsce]# ./runansible

Running local playbooks!

ERROR: There was an error while parsing the task 'lineinfile line=\'logformat combined %>a %ui %un [%{%d/%b/%Y:%H:%M:%S +0000}tl] "%rm %ru HTTP/%rv" %Hs %<st "%{Referer}>h" "%{User-Agent}>h" %Ss:%Sh\' dest=/etc/squid/squid-xs.conf'.

Make sure quotes are matched or escaped properly

PR #356, opened 2015-07-20
Moodlef2

georgejhunt:
earlier effort to enable password errors out and stalls a clean install.


jvonau:
Use runtime_php for the path

PR #357, opened 2015-07-21
download correct hostapd for architecture

georgejhunt:
I changed unleashkids to work with this for arm, i386, but not yet for x86_64


jvonau:
I understand the need with rpi2 but why not other arm proccessers and why
intel? Is this not in Fedora or is something broken? Trying to understand
the need for custom compiled software that needs long term support..

georgejhunt:
Good questions. I agree that we should only be substituting hostapd, when
the standard repo version does not meet our needs. In the two cases here,
I've had to compile a driver for the tplink wn725n, and the driver is not
compatible with stock hostapd.

And right now, that's only on rpi2 and xo1.5. I've changed the PR to what I
think is needed, but every time I look at it I see something I've missed.

PR #365, opened 2015-07-24
Use ansible_pkg_mgr in place of yum

jvonau:
used cmdline for the change: 
for files in \`egrep roles/\*/tasks/\*.yml -e yum: | awk -F'[:]' '{print $1}\`; do  sed -e "s|yum:|{{ansible_pkg_mgr }}|" -i $files; done

ansible_pkg_mgr exists in F18 should be safe for this to merge, yum groupinstalls have not been altered.


jvonau:
bad syntax

PR #366, opened 2015-07-28
Keep gw

jvonau:


jvonau:
corrects domain handling, speed improvements, preserves gateway without adding to bridge.

georgejhunt:
I was just about to start a testing phase on XO4, and Centos7.   I sort of
get gist of what your are doing, and it seems in the right direction --
simplify, simplify. and eliminate the stuck in lan_controller.

How do we want to test, as I think we're going into a QA phase. On
possibility


   1. run ./install_console
   2. in the console, just run with nothing additional enabled
   3. Then in the console, enable everything and run again again.
   4. Then from the cmdline ./run ansible (this will disable some things,
   but should not error out)
   5. Then add data, and get all the services that need data working --
   kalite, iiab, kiwix, rachel -- I'd like to automate from an external USB
   disk

Then for network configuration

   1. with one adapter ./runtags network -- set mode to appliance
   2. add second adapter, ./runtags network -- mode gateway
   3. remove gateway adapter, add wifi, -- mode lan_controller with adapter
   bridged
   4. add second adapter, leave wifi -- mode gateway
   5. remove second adapter, wifi -- mode appliance

PR #367, opened 2015-07-28
announce as http as well as https

georgejhunt:
we'll always have an http presence, even if the console switches to https


jvonau:
might want to copy it in place, think the schoolserver.service needs a
conditional on gui_port to set <type>_</type>

georgejhunt:
I think that Admin Console and Portal can be announced separately.

Tim has just the port variable changed via a console checkbox, but always
announces as _https._tcp.

This seems ok to me, since the real purpose is to make the ip address
visible.

PR #373, opened 2015-08-03
named is required with PEERDNS=no in ifcfg-WAN

jvonau:
push this if dns fails while in Appliance mode


georgejhunt:
So I pulled you fix, which puts 127.0.0.1 as dns.

But it turns out that named does not start  because the template for
/etc/named-xs has used xsce_domain="". --- and of course that is an error.
Line 66 is the first line it complains about.

jvonau:
Part of the speed increase is skipping what should be already completed, if
the domain didn't change after the first install I use that as a hint to
skip running playbooks, named is one of them.

The "xsce_domain" is used as defined in vars/*, I only set "xsce_domain" to
be the contents of "/etc/sysconfig/xs_domain_name" only if the contents are
not "lan". When "xsce_domain" has been set in vars a comparison between
"xs_domain_name" is done. If the same "xsce-prepped" is set, if they differ
xsce_domain will be used to overwrite what is currently in "xs_domain_name"
and "xsce-prepped" is not set giving you the full network run.

I still don't see how 'zone "{{ xsce_domain }}"' didn't get set in
named-xs.conf.j2

The fix is easy, delete /etc/sysconfig/xs_domain_name to force the full
network routine.

PR #379, opened 2015-08-05
Exfat

georgejhunt:
Our server needs to be able to read and write modern SD cards

### DIFF
--- a/roles/0-once/tasks/prep.yml
+++ b/roles/0-once/tasks/prep.yml
@@ -24,7 +24,7 @@
             src=li.nux.ro.repo
             owner=root
             mode=0666
-  when:     ansible_distribution == "Centos"
+  when:     ansible_distribution == "CentOS"
 
 - name: Create /etc/xsce
   file: path=/etc/xsce
--- a/roles/0-once/tasks/prep.yml
+++ b/roles/0-once/tasks/prep.yml
@@ -17,14 +17,14 @@
             src=rpmfusion-free-updates.repo
             owner=root
             mode=0666
-  when:     ansible_distribution == fedora
+  when:     ansible_distribution == "fedora"
 
 - name: Install proper repo -- for exfat
   template: dest=/etc/yum.repos.d/li.nux.ro.repo
             src=li.nux.ro.repo
             owner=root
             mode=0666
-  when:     ansible_distribution == Centos
+  when:     ansible_distribution == "Centos"
 
 - name: Create /etc/xsce
   file: path=/etc/xsce
--- a/roles/0-once/tasks/prep.yml
+++ b/roles/0-once/tasks/prep.yml
@@ -17,7 +17,7 @@
             src=rpmfusion-free-updates.repo
             owner=root
             mode=0666
-  when:     ansible_distribution == "fedora"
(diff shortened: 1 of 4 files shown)


jvonau:
I'd prefer if this was optionally opt-in given the restrictive licensing
noted.

https://en.wikipedia.org/wiki/ExFAT#Restrictive_licensing_and_software_patents

tim-moody:
http://superuser.com/questions/597431/cross-platform-usable-filesystem-with-large-file-support
If we don't have files bigger than 4g, we can use fat32, otherwise ntfs is a candidate.
Sent from my Samsung Galaxy smartphone.<div>
</div><div>
</div><div>-------- Original message --------</div><div>From: Jerry Vonau <notifications@github.com> </div><div>Date: 06/08/2015  2:04 PM  (GMT-05:00) </div><div>To: XSCE/xsce <xsce@noreply.github.com> </div><div>Subject: Re: [xsce] Exfat (#379) </div><div>
</div>
I'd prefer if this was optionally opt-in given the restrictive licensing
noted.

https://en.wikipedia.org/wiki/ExFAT#Restrictive_licensing_and_software_patents

On Wed, Aug 5, 2015 at 6:08 PM, georgejhunt <notifications@github.com>
wrote:

> Our server needs to be able to read and write modern SD cards
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/XSCE/xsce/pull/379
> Commit Summary
>
>    - add exfat drivers
>    - need quotes
>    - Fedora is capitalized
>
> File Changes
>
>    - *M* roles/0-once/tasks/prep.yml
>    <https://github.com/XSCE/xsce/pull/379/files#diff-0> (14)
>    - *A* roles/0-once/templates/li.nux.ro.repo
>    <https://github.com/XSCE/xsce/pull/379/files#diff-1> (8)
>    - *A* roles/0-once/templates/rpmfusion-free-updates.repo
>    <https://github.com/XSCE/xsce/pull/379/files#diff-2> (24)
>    - *M* roles/2-common/tasks/yum.yml
>    <https://github.com/XSCE/xsce/pull/379/files#diff-3> (2)
>
> Patch Links:
>
>    - https://github.com/XSCE/xsce/pull/379.patch
>    - https://github.com/XSCE/xsce/pull/379.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/XSCE/xsce/pull/379>.
>


---

georgejhunt:
I decided to use ntfs for my external passport -- takes twice as long to do
copies. So I'd prefer non-journalling FS for 128/256G SD cards. (maybe 4GB
max on file size is ok for a school server).

I think I can go with the flow regarding file format. In my process it's
just a few lines of code that need to change. And if we just download exfat
drivers, and let the user enable them, it seems like that might keep MS off
our backs. But concern about selling curated open source content in exfat
file format really  shoots down Adam's dream as I understand it.

I think I'd need to know more about MS's behavior to make a decision.

Thinking about installing exfat drivers, I thought I should start using
ansible generic package module. But it looks like they are not moving
toward generic very fast.  The generic module does not have a download vs
download and install (which is what we need for exfat to be optional).

PR #381, opened 2015-08-06
Vpn unit

georgejhunt:
This PR makes invocation of openvpn a systemd unit call rather than a rc.local event. It also points to an new cheaper and scaleable amazon instance. I've pulled it into master for syntax testing, but not been able to get past network problems to actually test function (openvpn comes late in ansible playbook)


jvonau:
{{ openvpn_server_port }} be set to the server's listening port 1194 not the ip address

jvonau:
Is the openvpn instance on xscenet.net active? nmap is showing port 1194 as filtered

georgejhunt:
On Fri, Aug 7, 2015 at 9:36 AM, Jerry Vonau <notifications@github.com>
wrote:

> Is the openvpn instance on xscenet.net active? nmap is showing port 1194
> as filtered
>
There is a filter. But the output of http://52.3.216.171/clients looks
valid

PR #382, opened 2015-08-07
prevent an observed failure

georgejhunt:
However it happened, I think this is the only place where xsce_domain is set, and I saw it = "".

### DIFF
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -6,7 +6,7 @@
 - name: Checking for prior domain name
   set_fact:
     xsce_domain: "{{ prior_domain.stdout }}"
-  when: prior_domain is defined and prior_domain.stdout != "lan"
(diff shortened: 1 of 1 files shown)


jvonau:
Yea, went for a walk and had another look, came to the same conclusion,
didn't think  prior_domain would register with the file no present, have a
look at PR#383.

tim-moody:
xsce_domain is set in default_vars and potentially in local_vars and config_vars.  I think this will mean that those values are never used except the first time.

jvonau:
So config_vars is not read all the time just when the gui is run so using
it's value as the final call is wrong also. The domain will remain the same
until xsce_domain  value is changed in one of the *_vars files.

tim-moody:
what does it mean to 'change' it?  if it is set to a value is that a change?

georgejhunt:
I think this problem is resolved with later PR's

PR #384, opened 2015-08-08
RFC move the default for xsce_domain to not be part of vars/*

jvonau:
Seems like the cleanest way of dealing with setting the default value while providing a means to change the domain on the fly. What concerns me is config_vars, the gui type-over box for domain appears to be reading config_vars while the system state might be different from the gui's last run. Think it would it be better to have that box be populated  with the current state when the page is first opened, saves the user from having to set domain in two places.

### DIFF
--- a/roles/1-prep/defaults/main.yml
+++ b/roles/1-prep/defaults/main.yml
@@ -2,6 +2,7 @@
 installing: False
 
 #Set defaults for discovery process
+old_domain: "none"
 discovered_wan_iface: "none"
 discovered_lan_iface: "none"
 discovered_wireless_iface: "none"
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -12,12 +12,12 @@
 - name: Setting prior domain name
   set_fact:
     xsce_domain: "{{ old_domain }}"
-  when: old_domain != "" and xsce_domain == ""
+  when: old_domain != "" and xsce_domain == "none"
 
 - name: Setting default XSCE domain name
   set_fact:
     xsce_domain: "lan"
-  when: xsce_domain == ""
+  when: xsce_domain == "none"
 
 # DETECT -- gateway and wireless
 - name: Get a list of slaves from previous config
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -19,7 +19,7 @@
 - name: Setting default XSCE domain name
   set_fact:
     xsce_domain: "lan"
-  when: old_domain == "" and xsce_domain == "none"
+  when: old_domain == "none" and xsce_domain == "none"
 
 # DETECT -- gateway and wireless
 - name: Get a list of slaves from previous config
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -6,8 +6,18 @@
 # above always registers
 - name: Checking for prior domain name
   set_fact:
-    xsce_domain: "{{ prior_domain.stdout }}"
-  when: prior_domain.stdout != "lan" and prior_domain.stdout != ""
+    old_domain: "{{ prior_domain.stdout }}"
+  when: prior_domain.stdout != ""
+
+- name: Setting prior domain name
+  set_fact:
+    xsce_domain: "{{ old_domain }}"
+  when: old_domain != "" and xsce_domain == ""
+
+- name: Setting default XSCE domain name
+  set_fact:
+    xsce_domain: "lan"
+  when: xsce_domain == ""
 
 # DETECT -- gateway and wireless
 - name: Get a list of slaves from previous config
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -13,7 +13,7 @@
  name: Setting prior domain name
   set_fact:
     xsce_domain: "{{ old_domain }}"
-  when: old_domain != "" and xsce_domain == "none"
+  when: old_domain != "none" and xsce_domain == "none"
 
 # First pass and no user override
 - name: Setting default XSCE domain name
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -206,6 +206,8 @@
             option='{{ item.option }}'
             value='{{ item.value }}'
   with_items:
+  - option: 'domain_name'
+    value: '{{ xsce_domain }}'
   - option: 'detected_gateway'
     value: '{{ discovered_wan_iface }}'
   - option: 'prior_gateway'
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -9,15 +9,17 @@
     old_domain: "{{ prior_domain.stdout }}"
   when: prior_domain.stdout != ""
 
-- name: Setting prior domain name
+# ok user didn't override default and xs_domain_name exists use it.
+ name: Setting prior domain name
   set_fact:
     xsce_domain: "{{ old_domain }}"
   when: old_domain != "" and xsce_domain == "none"
 
+# First pass and no user override
 - name: Setting default XSCE domain name
   set_fact:
     xsce_domain: "lan"
-  when: xsce_domain == "none"
+  when: old_domain == "" and xsce_domain == "none"
 
 # DETECT -- gateway and wireless
 - name: Get a list of slaves from previous config
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -1,11 +1,6 @@
 - name: Base Server Installed
   command: echo Base Server Installed
 
-- name: Restart xsce-cmdsrv service
-  service: name=xsce-cmdsrv
-           state=restarted
-  when: not installing
-
 - name: Restart httpd
   service: name=httpd
            state=restarted
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -147,6 +147,12 @@
   tags:
     - network
 
+#pre-populates xsce_domain correctly
+- name: Restart xsce-cmdsrv service
+  service: name=xsce-cmdsrv
+           state=restarted
+  when: not installing and not xsce_prepped
+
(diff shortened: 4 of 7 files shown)


jvonau:
Or do we need to record xsce_domain in xsce.ini to be read back into the gui?

georgejhunt:
a search of the ini file does not get a hit on the word "domain". 

But this thing has grown, and my aging brain cannot hold it all. 

This appears to deal with the error I found -- /sysconfig/xs_domain_name existing, but blank.

jvonau:
Yea I know, just wondering if we should record the domain in the xsce.ini
file.

I'm seeking info on how config_vars gets pre-populated, because config_vars
contains a value for xsce_domain, which may not be the same as what has
been defined in the past (think xs_domain_name), or currently defined in
local_vars.

We have a situation where the gui default for the domain type-over box is
defaulting to the value of the first install usually "lan", this value is
written in config_vars. Now config_vars is only used in the
xsce-from-console.yml playbook, which is read last after local_vars,
overriding all other setting.

If you don't change the domain in the gui to match any prior domain changes
whatever value that is in the box will get used even if you have defined
local_vars. I'm just looking how to grab the current state for the default
so we don't forget to or can't change it 2 places.

Just some thoughts.

jvonau:
Is this enough to display the current state of xs_domain_name in the "domain name" field in the gui?
or does index.html need some tweaking too?
I'm thinking on something like this, totally untested just looking for feedback.
----------------------------------
diff --git a/roles/xsce-admin/files/cmdsrv/xsce-cmdsrv b/roles/xsce-admin/files/cmdsrv/xsce-cmdsrv
index 0c7011e..21934d3 100644
--- a/roles/xsce-admin/files/cmdsrv/xsce-cmdsrv
+++ b/roles/xsce-admin/files/cmdsrv/xsce-cmdsrv
@@ -531,6 +531,7 @@ def cmd_handler(cmd_msg):
         "GET-VARS": get_install_vars,
         "GET-XSCE-INI": get_xsce_ini,
         "GET-CONF": get_config_vars,
+        "GET-DOMAIN": get_xsce_domain, 
         "SET-CONF": set_config_vars,
         "GET-MEM-INFO": get_mem_info,
         "GET-STORAGE-INFO": get_storage_info,
@@ -806,6 +807,14 @@ def get_zim_stat(cmd_info):
     resp = json.dumps(all_zims)
     return (resp)
 
+#we should always have a file, but.....
+def get_xsce_domain(cmd_info):
+    try:
+        resp = open(“/etc/sysconfig/xs_domain_name”).read()
+    except IOError:
+        resp = "lan"
+    return (resp)
+
 def get_config_vars(cmd_info):
     read_config_vars()
     resp = json.dumps(config_vars)
diff --git a/roles/xsce-admin/files/console/js/admin_console.js b/roles/xsce-admin/files/console/js/admin_console.js
index e4233e2..687d814 100644
--- a/roles/xsce-admin/files/console/js/admin_console.js
+++ b/roles/xsce-admin/files/console/js/admin_console.js
@@ -277,6 +277,11 @@ function button_feedback(id, grey_out) {
   }
 }
 
+function find_current_xsce_domain()
+{
+   xsce_domain = get_current_xsce_domain()
+}
+
 // Field Validations
 
 function xsce_domainVal()
@@ -491,6 +496,7 @@ function setRadioButton(name, value){
 function initConfigVars()
 {
   assignConfigVars();
+  find_current_xsce_domain();
   var html = "Gateway: ";
   if(typeof ansibleFacts.ansible_default_ipv4.address === 'undefined'){
     html += "Not Found<BR>";
@@ -1249,6 +1255,13 @@ function getHelp(arg)
   return true;
 }
 
+function get_current_xsce_domain()
+{
+  command = "GET-DOMAIN"
+  sendCmdSrvCmd(command, genericCmdHandler);
+  return true;
+}
+
 function formCommand(cmd_verb, args_name, args_obj)
 {
   cmd_args = {}

---------------------

georgejhunt:
I think if you made a PR, we could use github to display python in a more
readable format. I cannot get a sense of it.

jvonau:
Sorry George, I replaced the patch with a link, but I found an easier way.... Please view the PR
https://github.com/XSCE/xsce/pull/384

georgejhunt:
I like the domain setting.

But I'm confused.  An earlier email had a lot of poorly formatted
javascript that is not in the PR.

jvonau:
Yup replaced that munged up patch (pasting into github not nice... lol)
with a link in the PR to the patch. That patch is not required, cmdsrv runs
ansible-setup when the service first starts not when logging into the gui
webpage, so restart cmdsrv in network if the domain changed. This is fine
when using install-console, runansible, or runtags, I'm not sure if
restarting cmdsrv is the right thing to do when running from the console
thou.... Thoughts?

georgejhunt:
It might even be a two step process. First the cmd-srv needs to start over
filling its dictionaries. Then the page needs to request a refresh, and get
a new dump of data -- that's my limited understanding.

I think we should wait for Tim's to respond too.

tim-moody:
The real question here is how to configure and maintain the server.  Should it be done with unix commands, ansible, or a gui.  The immediate focus is setting the domain name and where is the definitive value to be used for setting it.

 

My view is that if someone wants to set up a server using commands they can do so.  No need to get the xsce repo, just install packages you want, configure them, and you're done.

 

If you want to use ansible, the values for configuration should come from ansible vars.  Clone the xsce repo and use ansible to install and configure the server.  If you go back to using commands, you are on your own.

 

If you want to use a gui, ansible installs one.  Use it to configure and install things.  If you go back to using commands, you are on your own.

 

So ansible vars win over things set with the command line and gui vars win over default and local vars.

 

Again in my view xs_domain_name is not a configuration, but a record of what has been set by ansible for use by other applications outside of installation and configuration.  The value to use for setting the domain is that in the hierarchy of default, local, and config vars.

jvonau:
On Sun, Aug 9, 2015 at 1:20 PM, Tim Moody <notifications@github.com> wrote:

> The real question here is how to configure and maintain the server. Should
> it be done with unix commands, ansible, or a gui. The immediate focus is
> setting the domain name and where is the definitive value to be used for
> setting it.
>
>
>
Ok let's discuss it.


>
> My view is that if someone wants to set up a server using commands they
> can do so. No need to get the xsce repo, just install packages you want,
> configure them, and you're done.
>
>
>
What does that have to do with this discussion then?


>
> If you want to use ansible, the values for configuration should come from
> ansible vars. Clone the xsce repo and use ansible to install and configure
> the server.


Ok, the instruction suggest that both ./install-console and ./runansible
are being supported.


> If you go back to using commands, you are on your own.
>
>
What commands are you referring to? Using runtags? git pull? nano? vim?


>
>
> If you want to use a gui, ansible installs one.


One has no choice in that, but it does support more optional services.


> Use it to configure and install things. If you go back to using commands,
> you are on your own.
>
>
That sounds a bit like empire building.... :) I don't see why both methods
can't be used interchangeably. I don't want to have to tell a user I can't
fix their mess because the gui was used. lol..
Can't recommend a git pull for the latest update/new functionality because
local_vars will be toasted, losing ansible's only reference to a non-stock
domain name in cli or when config_vars is {}. Remembered that one by
forgetting to rename local_vars, brought it back to the top of my head.


>
>
> So ansible vars win over things set with the command line and gui vars win
> over default and local vars.
>
>
>
The gui might suffer the same fate as the cli in respect to xsce_domain
should one fail to log into the gui then hit "save config" right after
installing to populate config_vars, that fixes the disappearing local_vars
with xsce_domain set problem for the gui.
Maybe add that to the release notes, best practices?

The gui displaying stale data is more of an issue after first install if
xsce_domain changed in local_vars then access the gui without restarting
cmdserv. Better add that to the release notes, then I'll feel a bit better
about the potential stale data in the gui. I could restart cmdserv in
runtags. Think I like that, should help keep things in sync between gui and
cli.


>
> Again in my view xs_domain_name is not a configuration, but a record of
> what has been set by ansible for use by other applications outside of
> installation and configuration.


Yes I fully agree, but there is no valid reason I can think of not to let
ansible itself reference the file for information and use it for logical
decisions. I don't want to rely on local_var to set the domain name for
ansible after the first install, so in the future we might be able to do in
the field updates with less fuss.


> The value to use for setting the domain is that in the hierarchy of
> default, local, and config vars.
>
>
I'd prefer to have local last, given it is empty by default and intended as
a user override but whatever, I can live with it.

I'm going to clean this PR up and summit a revised one.

But the original question sort of was "how bad is it to restart cmdserv
while ansible is running under it?"

tim-moody:
Adding this manually as reply didn't work:

My first question is if the host's domain is not the same as xsce_domain, how did it get that way and what should be done about it.  My view is it should be set to the value in  xsce_domain as defined in the chain of var files, not detected.  I assume it got that way either because no install has ever been done and it is what was set automatically or manually when linux was installed or someone used a linux command to change it.  In either case ansible vars win.

Second point is that install-console sets local_vars to non-track so that it will never be clobbered by a git pull.  I don't think config_vars can get clobbered either.  For my xsce.lan install I have my own local_vars that I copy from xsce-local to xsce.  It is never clobbered by a pull.

I attempted to document all this in https://github.com/XSCE/xsce/blob/master/docs/VARIABLES.rst quite some time ago.  We have too many sets of instructions, written by different  people and they are inconsistent and often out of date. 

I agree that switching between manually running runansible or runtags and using the console is messy.  But I thought we discussed this (maybe only with George).  Are you happy with the console clobbering local_vars?  I wasn't.  In the worst case someone could be editing the file while someone in the console was also changing it.  I don't think changing the order of precedence helps because now the gui use keeps changing some that never changes.

My view is that both runansible and console are supported, but not interchangeably.  If you're a power user and want to use runansible/tags do so, but then if you switch to console, stick with it.

If the gui is now displaying stale domain data, I don't know why, unless by stale you mean that it was changed in local_vars after using the gui.  If you change the domain via the gui does the gui see the change?

If the original question was how bad is it to restart cmdsrv if it is running ansible the answer I think is that it will kill ansible as it is a subproc.  But another answer would be to add a command to refresh data.  Commands can be submitted from the command line with xsce-cmdsrv-ctl <command>.  My assumption as you can see above is that this is never necessary, because the cmdsrv always has the correct version of the data including xsce_domain.

Finally, if there are optional services that can't be managed with the gui, that should only be due to the fact that they haven't yet been added and they should be.  I agree with your position on teamviewer and would like to add buttons to start and stop including xwindows.  Probably openvpn should work the same.

PR #387, opened 2015-08-10
Replacement for PR #384

jvonau:
Replacement for PR #384

### DIFF
--- a/roles/1-prep/defaults/main.yml
+++ b/roles/1-prep/defaults/main.yml
@@ -2,6 +2,7 @@
 installing: False
 
 #Set defaults for discovery process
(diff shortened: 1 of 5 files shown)


georgejhunt:
I like the speedup provided by skippiing stuff when the domain name has not changed. But I wonder if there are not corner cases when "./runtags network" shouldn't go through the whole process, and not skip anything. And I don't know how to sense the "./runtags network" situation. 

Maybe we redefine the old xs-setup-network script, and have it do:

  * rm -f /etc/sysconfig/xs_domain_name
  * /opt/schoolserver/xsce/runtags network

PR #389, opened 2015-08-14
Reconcile install

tim-moody:
create install-init script and playbook to set up var files before subsequent scripts
call install-init at the beginning of runansbile, runtags, and install-console
change top level playbooks to use local_vars from /etc/xsce
change top level playbooks to add config_vars from /etc/xsce
fix a few syntax errors in comments
remove some trailing blanks

### DIFF
--- a/docs/VARIABLES.rst
+++ b/docs/VARIABLES.rst
@@ -17,27 +17,29 @@ Variable Files
 
 The vars directory holds two files, default_vars.yml and local_vars.yml.  The former holds
 the default values for a number of global variables for the installation.  The latter allows
-deployments to override these values.  Changes should not be made to default_vars.yml as they could
-be overwritten by a subsequent pull from the git repository.  The inital local_vars.yml file comes
-from the git repo but is marked not tracked on the first run, so edits will not be lost.  These
-parameters are either ones that a deployment may wish to override, such as xsce_domain, or global constants
-such as xsce_config_file.
+deployments to override these values.  These parameters are either ones that a deployment
+may wish to override, such as xsce_domain, or global constants.
+
+The inital local_vars.yml file comes from the git repo but is marked not tracked on the first run,
+so edits will not be lost, and is copied to /etc/xsce/local_vars.yml where it may be edited.
+Changes should not be made to default_vars.yml as they could be overwritten by a subsequent
+pull from the git repository.
 
 Variables Set in the Admin Console
 ==================================
 
-When installation is carried out using the Admin Console an additional variable file is used, /etc/xsce/config_vars.yml,
-which is set via a graphical user interface.  Values in this file further override values in default_vars.yml and 
-local_vars.yml files. It should be kept in mind that no changes are made to the original local_vars.yml file, so that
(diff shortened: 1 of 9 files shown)


georgejhunt:
So as I understand it, once a user has run any config from the console, he
needs to copy /etc/xsce/console.yml to /etc/xsce/local_vars.yml, before he
does ./runtags, or ./runansible.

If that is so, maybe we need an easy to remember script name like
"revert_to_commandline" which does "cp /etc/xsce/console_vars.yml
/etc/xsce/local_vars.yml". Or maybe we just initialize /etc/xsce/local_vars
with what console_vars.yml is currently initialized with. Then as users, we
can look at a cmdline/console variable which we can check to see who is in
command.

Probably I don't understand enough. Let's talk on Saturday, at Adam's 10am
EDT meeting.

Or maybe  there's a reminder in runtags (I'm thinking of the "./runtags
network" situation) which checks a flag to see whether console config has
ever been run (I think in the old situation, console_vars.yml did not
exists until console config had occurred).

Until the console is more complete, there will always be things that can
only be done at the command line. We'll need some way, doing remote
administration without a GUI, to tweak things without destroying the
integrity of the console displayed data.

tim-moody:
No; not any more.  Config_vars is now included in all the scripts, so there is no reason to copy anything.  But if you want to set a value, you should use the console, even if you like running runtags manually.

 

Things that the console does not configure can still be configured with local vars, but config vars wins over local vars for things that can be set either place, because the reverse will confuse gui users.

 

 

So as I understand it, once a user has run any config from the console, he
needs to copy /etc/xsce/console.yml to /etc/xsce/local_vars.yml, before he
does ./runtags, or ./runansible.

Or maybe there's a reminder in runtags (I'm thinking of the "./runtags
network" situation) which checks a flag to see whether console config has
ever been run (I think in the old situation, console_vars.yml did not
exists until console config had occurred).

 

yes, but now it does.

Until the console is more complete, there will always be things that can
only be done at the command line. We'll need some way, doing remote
administration without a GUI, to tweak things without destroying the
integrity of the console displayed data.

georgejhunt:
Could we eliminate local_vars.yml? And always just use config_vars.yml?
Then we would not need to even think about whether console knows about it
or not.

tim-moody:
I take it you mean having the console edit a file that can also be edited with a text editor.  I thought about this, but it is harder than it might seem.  If you read a yml file in python you lose all the comments.   There is a ruamel library, but it converts lines into a special object, so that they are no longer lists or dicts and so harder to work with.  Also, if someone messes up a line the console will be broken.  There needs to be a home for things like password hashes that are not yet or possibly ever handled in the console, things that might span installs like the country level xsce-admin password.

georgejhunt:
Or just always automatically append/replace console_config.yml to the end
of local_vars, such that it always overrides local_vars, and is visible to
cmdline folks who are wondering why their changes are not honored (or get
overridden by GUI user tomorrow, or next year).

I'm only trying for consistent, understandable behavior.

PR #392, opened 2015-08-17
The include_vars module does not like an empty vars file (all comment…

georgejhunt:
…s), so add a dummy variable to local_vars


tim-moody:
sorry; I guess my test file had data.

PR #393, opened 2015-08-17
0-once & local_vars handling

jvonau:
Proposed revision to docs to reflect the change in local_vars handling. Avoid duplication in running playbooks.


tim-moody:
I missed this in cmdsrv if local_vars is truly moved to /etc/xsce

1366
stream = open(xsce_git_path + "/vars/local_vars.yml", 'r')
    local_vars = yaml.load(stream)
    stream.close()

PR #396, opened 2015-08-19
Gui Gateway -> Appliance

jvonau:


georgejhunt:
Is this ready for me to test? I thought there was a xsce_lan_enabled which had to be set false by the gui cmdsrv routine.

jvonau:
With this there would not be any gui changes needed.

tim-moody:
In case this didn't answer the question and for clarity.

 

the gui sets gui_desired_network_role to one of appl, gw, or netc (not literally)

 

the rest of the story is Jerry's dept

 

this value should not be used except by computed_network.yml (I didn't find it there!)

 

to decide what topology to configure.

tim-moody:
Looked again and gui_desired_network_role is indeed used by computed_network to override gateway with appliance.  I understand that checking gateway doesn't make sense as it is either already set or not able to be set, though not sure of the interactions with local vars.  If local vars said to prefer appliance but gui said gw and gw is possible, then it should be gw.

Similarly, if the gui user asked for NW contr and the thing is set to appliance, are we not able to accommodate that wish?

jvonau:
Want to discuss some proposed default behavior for when user wants network
modes change done in the GUI.

On Thu, Aug 20, 2015 at 8:20 AM, Tim Moody <notifications@github.com> wrote:

> Looked again and gui_desired_network_role is indeed used by
> computed_network to override gateway with appliance. I understand that
> checking gateway doesn't make sense as it is either already set or not able
> to be set, though not sure of the interactions with local vars. If local
> vars said to prefer appliance but gui said gw and gw is possible, then it
> should be gw.
>
How would one "prefer appliance" when vars/default_vars doesn't have
gui_desired_network_role or any mention of network role. What is available
is xsce_lan_enabled which could be set to False in /etc/xsce/local_vars to
achieve that effect

Now with config_vars winning out we have to define what the other settings
will do by default.
Appliance -> Gateway = add files always worked, done
Gateway -> Appliance = Delete all LAN info, now done

> Similarly, if the gui user asked for NW contr and the thing is set to
> appliance, are we not able to accommodate that wish?
>
These two, we need to know if 1) we should blow away the gateway and become
part of LAN or 2) prevent the old gateway interface from being considered
part of LAN and be "reserved" for gateway use again.
Gateway -> LanController
Appliance -> LanController Note, if there is no other interface don't try
to reserve gateway.

Breaking out of LanController becomes more difficult when no interface is
reserved.
LanController -> anything

When 2 or more interfaces are in the LAN, we have to take down the LAN
slaves and test one by one.  I'm in favor of providing a shortcut for the
user, a field for the wan device name in the gui which is in effect going
to become user_wan_iface directly or entering/defaulting to "AUTO" will
kick off the re-detection outlined in the other emails. The re-detect is
going to be intrusive and I want to be able to avoid it if at all possible
for ssh reasons but should be fine if run from the gui.

I hope that helps clear up the issues as I see them,


Jerry

tim-moody:
Yes, on call tomorrow.  But I don't think of it quite like that.  I think the gui user says what they want pure and simple without necessarily knowing what is there or will be there in the future.

 

My question was to explore what would happen if local-vars had xsce_wan_enabled: False and config_vars had gui_desired_network_role: Gateway.

 

Networking is the only place where there are parallel sets of variables.

PR #398, opened 2015-08-19
Wifi

georgejhunt:
eliminate testing confusion with multiple wifi units on my desk

### DIFF
--- a/roles/network/defaults/main.yml
+++ b/roles/network/defaults/main.yml
@@ -6,3 +6,4 @@ wondershaper_dspeed: "4096"
 wondershaper_upspeed: "1024"
 ssid: "XSCE"
 wifi_mode: g
+channel: 6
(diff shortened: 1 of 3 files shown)


tim-moody:
actual values for a deployment would go in local_vars, right?
which wifi does this refer to, internal WAN, internal LAN AP, external AP?

tim-moody:
what about password?

georgejhunt:
This refers to hostapd, which is the internal lan. Yes local_vars is where
this would be set

For the internal wan on an xo1.5 we can continue to use gui  from the OLPC
network neighborhood.

External AP can be administered by whatever AP is used, and it's normal
config process

georgejhunt:
Password for our internal AP is complicated by setting up wpa_supplicant.
Maybe it will be necessary long term, but open seems ok to me for this
release.

tim-moody:
could the variables be named to indicate that this is hostap_ssid, etc.  we might one day have wan_ssid

georgejhunt:
I guess I should spend the time now to work our psk2 passwords for hostapd.
I'll put it on my list.

PR #402, opened 2015-08-20
Atboot

georgejhunt:
The PR seems like a good vehicle to propose, and hash over ideas, maybe better than email


tim-moody:
superceded

PR #406, opened 2015-08-22
Install doc

jvonau:
Needs some formatting to look nice on the web but it is better than nothing....


georgejhunt:
Thanks for the additional documentation. When I tried to use your
installer, I didn't have some needed information, and I wasn't very
successful.

Does the installer, as it is currently, do partitioning that includes swap,
boot, root, and library?

Given that FC23 is due out Nov 2015, and most of the FC22 bugs that will be
fixed are probably updated out, maybe we should bump to FC22.

I think all of Tim's testing has been on FC22 VM.

jvonau:
Not sure what happened to my email response but here it is again:
On Sat, Aug 22, 2015 at 8:46 AM, georgejhunt <notifications@github.com> wrote:

    Thanks for the additional documentation. When I tried to use your
    installer, I didn't have some needed information, and I wasn't very
    successful.


The livecd method is OK but having to build and host files creates work for us, that I want to avoid. Add to that fedora's messing with "fedora-release-*" made the livecd's more difficult to produce for F21+, but I didn't throw much time at it knowing this alternative.
 

    Does the installer, as it is currently, do partitioning that includes swap,
    boot, root, and library?


You're in the stock fedora installer, with the installer guided by the kickstart file.  The partitioning is not automatic out of the box but could be once we know the correct syntax that kickstart needs for what we want to do.

 

    Given that FC23 is due out Nov 2015, and most of the FC22 bugs that will be
    fixed are probably updated out, maybe we should bump to FC22.

    I think all of Tim's testing has been on FC22 VM.

 
Run this with F22/F23's net-install boot. iso and boot the resulting key, should just work as the kickstart was written to be release agnostic.
CentOS will need a different kickstart file with a bit of work to enable the epel repo to be able to install ansible during install time.


If you try this the /xsce/*.log files would tell the tale, and I could help interpret them should things go less than smooth.

PR #407, opened 2015-08-22
RFC console speed up when networking has not changed.

jvonau:
We can skip the network playbook if the detected network layout matches the user's wishes in the gui.

### DIFF
--- a/roles/3-base-server/meta/main.yml
+++ b/roles/3-base-server/meta/main.yml
@@ -2,4 +2,4 @@ dependencies:
    - { role: httpd, tags: ['services','httpd','base'] }
    - { role: portal, tags: ['services','portal','base'] }
    - { role: xsce-admin, tags: ['services','xsce-admin','base'] }
-   - { role: network, tags: ['services','base','network'] }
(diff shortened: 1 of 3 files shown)


tim-moody:
I think gui_desired_network_role will be defined as soon as the console has been run once

jvonau:
That would be the key, if user wants a certain layout and that matches what is detected I see no point in running through the playbook to make no changes, otherwise run the playbook.

tim-moody:
agreed.  is that what - { role: network, tags: ['services','base','network'], when: not gui_desired_network_role is defined } does?

jvonau:
On Sat, Aug 22, 2015 at 12:01 PM, Tim Moody <notifications@github.com>
wrote:

> agreed. is that what - { role: network, tags:
> ['services','base','network'], when: not gui_desired_network_role is
> defined } does?
>
>
Untested,  I was looking to gather thoughts before proceeding. I'm hoping
that it does and will report back in this PR.

jvonau:
part of PR #510

PR #412, opened 2015-08-22
kalite - don't try to git clone twice

jvonau:
don't try to git clone twice, observed failure where the git repo is already present.
templates to install, could stay if we're looking to change port numbers on the fly.
run yum install based on data.sqlite's presence, for some speed

### DIFF
--- a/roles/kalite/tasks/enable.yml
+++ b/roles/kalite/tasks/enable.yml
@@ -29,7 +29,8 @@
             section=kalite
             option='{{ item.option }}'
             value='{{ item.value }}'
-    - option: enabled
-      value: "{{ kalite_enabled }}"
-    - option: cron_enabled
-      value: "{{ kalite_cron_enabled }}"
+  with_items:
+    - option: 'enabled'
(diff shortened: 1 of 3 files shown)


tim-moody:
Just so you're clear, if any of the following fails it will never get rerun (currently true as well)

- name: Run the install using kalite manage
  command: "/usr/bin/su {{ kalite_user }} -c '{{ kalite_root }}/bin/kalite manage setup --username={{ kalite_user }} --password={{ kalite_password }} --noinput'"
  async: 300
  poll: 10

- name: Finish install by running kalite start
  command: "/usr/bin/su {{ kalite_user }} -c '{{ kalite_root }}/bin/kalite start'"
  async: 300
  poll: 10

- name: Stop kalite server started in previous step because we use systemd
  command: "/usr/bin/su {{ kalite_user }} -c '{{ kalite_root }}/bin/kalite stop'"

jvonau:
I'm OK with that, I haven't seen a setup failure in a long while. Good job

jvonau:
superseded by later changes

PR #415, opened 2015-08-23
stat detect not working as intended

georgejhunt:
Jerry,

It's a long story. 
I forgot that my AP with ssid:"George's Time Capsule" gives our code fits because of the apostrophe. NM created an ifcfg for it, and runtags network burped on the apostrophe.

So I deleted the ifcfg file, associated with another AP, and rantags network. The comments suggest that checking for a file was your intent, but it did not work because the variable you tested was not defined.

Hence this PR. I can help chip at the tests, even if I cannot see the larger picture

### DIFF
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -26,7 +26,7 @@
 - name: Setting has_WAN
   set_fact:
     has_WAN: "True"
-  when: has_ifcfg_WAN.stat.exists
+  when: has_ifcfg_WAN.stat.exists is defined and has_ifcfg_WAN.stat.exists
 
 # unused
 # DETECT -- gateway and wireless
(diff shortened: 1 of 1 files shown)


jvonau:
Thanks, I'm just looking at that too... This was working just fine.. but now after peppering the code with debug and pause I see that:
1. has_WAN: hasn't defaulted to False from default/main even before setting anything and it should this concerns me as we have a bunch of default... 
2. the stat is not triggered when the file is present, noted by - debug: msg="{{has_ifcfg_WAN}}"
I've hand tried you change with no luck here
Can you check that your change is setting has_WAN by adding just after " Setting has_WAN" 
- debug: msg="{{has_WAN}}"
- pause:

georgejhunt:
I tried the debug, when ifcfg-WAN existed, and then I deleted it, and got
correct results in each case.

jvonau:
Momentary brain fart on my part, must of been tired when adding debugs, had a space after=, which tainted what I was seeing with has_WAN. The has_WAN works as is without mods, but ifcfg_gw_device should be using stdout_lines, changed in 29b4189ec1b6ed3a6b80613a52ff85cf65732ae3 part of PR #416

PR #439, opened 2015-09-14
Check lan

jvonau:


georgejhunt:
How is this kicked off? Could it be tacked on to a finish install unit file?

jvonau:
Just call the file, logs created in systemd's journal and /var/log/messages, intended to be called by a unit file

PR #445, opened 2015-09-15
need ncat to report user handle

georgejhunt:


jvonau:
More coming or just push it?

PR #472, opened 2015-10-15
WordPress

tim-moody:
playbook to install wordpress as part of 6-generic
soft code home page path (can be /xs-portal, /wordpress, /dokuwiki, but they must be installed)
wp install page comes up on first run to enter administrator user name and password (in future we will populate during the install as Admin/changeme)

access WordPress at http://schoolserver[.lan] (or whatever works)/wordpress.


tim-moody:
can't ever get urls to work in github.  the above should be whatever url works to reach the schoolserver plus /wordpress (slash wordpress)

jvonau:
#xsce_home_url: /wordpress
#xsce_home_url: /dokuwiki

Just a thought, maybe have some logic for xsce_home_url:<option> to auto set <option>'s install and enabled facts? Just so the users don't have to remember to set the install & enabled options in a vars file, it could be in just one place to select and enable the requested backend like the db backends are done in  computed_vars.yml.

tim-moody:
I'd like to hold off on such logic until it settles down and we get it into the console, but in general, the interaction between enabled and portal selection has to be considered.  At the same time you could decide to use xs-portal for your portal, wp for your blog, and dokuwiki for collaboration and want them all enabled.

PR #477, opened 2015-10-22
enable/disable phpmyadmin on phpmyadmin_enabled

tim-moody:


tim-moody:
handle the default disabled
do the gui later

PR #480, opened 2015-10-29
Kalite 0.15

tim-moody:
substantially revised to use 0.15 except on f18 which is still 0.13.2
tested on x86_64 vm on f22
please test on xo and centos


tim-moody:
I ran the new kalite playbook on an xo-1.5 (installs 0.13.2).  It works more or less the way the old one did, which is to say

1) the --recursive on the git problem errors out on first run
2) the setup can time out, even at 900 seconds
3) it is all very fussy. it took three tries to complete the install (and one rm of /library/ka-lite), but in the end it works.

on f22 vm the playbook installs 0.15 and I think it works.

PR #483, opened 2015-11-06
GET-STORAGE-INFO

tim-moody:
rewrite to accommodate inconsistencies in blkid and by-uuid across versions of os

use lsblk instead


tim-moody:
tested on xo 1.5 f18 and 64 bit nuc f22

PR #487, opened 2015-11-12
Net nov12 - feedback please

jvonau:
3 points 
1. redetect.yml needs edit_ifcfg.yml to rename has_ifcfg_gw to be xsce-WAN. think0 on my part.
2. call nmcli c reload to clear out deleted files from memory before re-detection. should make this part much more robust.
3. user_wan_iface needs to be auto for redetect.yml to occur. Might be better to alter 1-prep for  gui_wan_iface?

### DIFF
--- a/roles/network/tasks/computed_network.yml
+++ b/roles/network/tasks/computed_network.yml
@@ -30,10 +30,10 @@
 # we need to have an interface name for ifcfg-WAN to be able to change gateway
 # the DEVICE from the gui. Thanks to George for proving my point about knowing
 # what device to switch to.
-- name: Using GUI_WAN info
-  set_fact:
-    user_wan_iface: "{{ gui_wan_iface }}"
-  when: gui_wan_iface != "unset" and gui_desired_network_role is defined and gui_desired_network_role != "LanController"
+#- name: Using GUI_WAN info
+#  set_fact:
+#    user_wan_iface: "{{ gui_wan_iface }}"
+#  when: gui_wan_iface != "unset" and gui_desired_network_role is defined and gui_desired_network_role != "LanController"
 
 # should make the GUI buttons the last call
 - name: Checking xsce_wan_enabled
--- a/roles/network/tasks/ifcfg_mods.yml
+++ b/roles/network/tasks/ifcfg_mods.yml
@@ -60,7 +60,7 @@
 
 # test point, we should always have one with any kind of starting point
 # no ifcfg = supply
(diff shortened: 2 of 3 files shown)


tim-moody:
I'm not clear what this PR fixes

when: has_wifi_gw == "none" and has_ifcfg_gw != "none" and has_ifcfg_gw != "/etc/sysconfig/network-scripts/ifcfg-WAN"

the names imply boolean, but the values are otherwise

shell: ls /sys/class/net | grep -v -e lo -e br - should also exclude tun?

gui_wan_iface != "unset" - where is this value set?

jvonau:
On Sun, Nov 15, 2015 at 10:08 AM, Tim Moody <notifications@github.com>
wrote:

> I'm not clear what this PR fixes
>
> when: has_wifi_gw == "none" and has_ifcfg_gw != "none" and has_ifcfg_gw !=
> "/etc/sysconfig/network-scripts/ifcfg-WAN"
>
> the names imply boolean, but the values are otherwise
>
Would removing "has_" from the variable names make things clearer for you?


> shell: ls /sys/class/net | grep -v -e lo -e br - should also exclude tun?
>
That wouldn't hurt, just to be sure.

> gui_wan_iface != "unset" - where is this value set?
>
network/defaults/main.yml is the default but might be set in
detected_network in 1-prep:

# OK try old gw this is a best guess based on what's in
# /etc/sysconfig/xs_wan_device's last state intended to
# provide a seed value to display in the GUI when no
# gateway is present but we had one.
- name: Has old gateway and no discovered gateway setting WAN
  set_fact:
    gui_wan_iface: "{{ device_gw }}"
  when: user_wan_iface == "auto" and device_gw != "none" and
discovered_wan_iface == "none"

That is a cheap workaround for the GUI's lack of the input field. Kind of
need to know which interface to apply the static addressing to. cmdline
vars you could set user_wan_iface but there is no equal in the GUI.

Setting user_wan_iface to anything but 'auto' disables redetect.yml's try
dhcp on all interfaces use the one device that responses. I'm on the fence
on whether it's better to disable gui_wan_iface in 1-prep/detected_network
or network/computed_network.

georgejhunt:
I'm of the opinion that it's too late in our release cycle to be changing
anything related to networking that is not in response to a specific
failure (maybe this is that, and I'm just not following closely enough).

I'm just remembering all the "nmcli dev eth0 up/down's" that errored out on
a different platform than the testing/development was done on.

PR #488, opened 2015-11-14
Fix usbmount early, before GUI is needed

jvonau:
should result in the fix being applied during install-console and not needing a full run to become active


tim-moody:
I think if you are going to create the usb-mount section of the ini, then the desc should also be written

jvonau:
Or just remove the recording of the alteration, we don't record other
similar modifications like chrony, sysctl, or sshd.

tim-moody:
agreed.  just record it later in the usb_mount role with the other parameters

PR #491, opened 2015-11-17
Centos xovis

georgejhunt:
Maybe we need to document that xovis is not available on CentOS, because it might swing some people to FC


jvonau:
using meta works but not as clear as moving 'the guts' to an install.yml file conditionally called from main.yml like how kalite is handled

jvonau:
I don't think the second commit is needed, there is code just like the first commit for sugar-stats where it works skipping the playbook while vars can be 'True'

PR #493, opened 2015-11-18
do not merge RFC couple of ideas to toss about

jvonau:


tim-moody:
can we look for /xsce-update instead of /schoolserver

tim-moody:
some of us use xsce-local for customization.
check if /opt/schoolserver/xsce-local exists and if so
run /opt/schoolserver/xsce-local/scripts/cp-ans before running
and /opt/schoolserver/xsce-local/scripts/cp-root after

tim-moody:
are there limits on which directories can be updated?

tim-moody:
my strategy was to do this sort of thing through the gui under user control and consent

jvonau:
/xsce-update can be set at MAGIC_DIR
I'd have to look at what xsce-local does, but that should be do-able.
No limits, reads the directory listing of MAGIC_DIR, what is done with with
each is up to the case statement
handling.
I think it would be safer to update the gui from outside rather than within
so only that is updated. Other updates after updating would need an ansible
run to become active.

I'll take having the presence of MAGIC_DIR as implied consent, just like
the auto activity handling. We could check for a flag file, manifest...
it's up for debate.

Note the introduction of a version number(git tagging or just a file in
git? just a number is easier) for the state of things done during
install-console.

davelab6:
@jvonau perhaps rebase (interactive) to fixup the 2nd commit?

davelab6:
This seems quite old though, perhaps should be closed?

jvonau:
closing

PR #494, opened 2015-11-18
install-init changes

tim-moody:
only run if /etc/xsce/config_vars.yml doesn't exist
add don't track (used to be here, but disappeared sometime)

### DIFF
--- a/install-init
+++ b/install-init
@@ -1,9 +1,5 @@
 #!/bin/bash
 
-# Mark local_vars as do not track to preserve local changes
-
-git update-index --assume-unchanged vars/local_vars.yml
-
 # the playbook does nothing if the following already exists so exit
 
 if [ -f /etc/xsce/config_vars.yml ]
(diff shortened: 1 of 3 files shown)


jvonau:
That git call has always been in local_facts.fact.j2, is only run during 0-once, but a git reset might undo that anyways. Remove from  local_facts.fact.j2? . I like having the call in install-init. Once this is pushed I have to fix the the kickstart files logic, I'm looking for the results in the logs and was testing the number of failed=0 that are present.

tim-moody:
I thought it was somewhere and I probably put it there; it's my style of comment.
I guess what happened is that I did a clone and cp from xsce-local, but later did a pull without ever running install-console or any other script so local_vars got overwritten.  This change wouldn't fix that, but would allow install-init to be run to create an image without anything being installed.

tim-moody:
maybe this belongs in the xsce-local cp-ans script.

jvonau:
think "cd $XSCE_DIR" is still needed to put you in the directory for the git command

jvonau:
I think "if [ ! -f xsce.yml ]" should be install-init.yml

jvonau:
Ok, the cd is not needed, we agreeded not to run ansible from outside of $XSCE_DIR and we are in the git repo when running.

PR #496, opened 2015-11-19
restrict udev patch to Fedora

tim-moody:


tim-moody:
What if anything needs to be done on Centos?

I assume this failed because "Centos" is >= "21"

PR #502, opened 2015-11-21
use wget instead of get url to avoid out of space on /tmp

tim-moody:


georgejhunt:
I don't think this will work:

chdir: somedir/+ creates: /path/to/database+ Though I could be missing
something

tim-moody:
yep, tested on one machine and edited on another.  left the sample code in the playbook.

PR #517, opened 2015-12-01
Adams install-console fix

jvonau:
This point of the code was designed to be used early to bring up the network before yum was run, and may be out of place in the scheme of things now. For this to really work as intended network would need to move to just after 1-prep in install-console. For now just skip supplying the WAN file at this point as there is additional logic later dealing with gui_static_wan in computed_network.yml


jvonau:
I would be ok with this block of code just being removed as the liveiso/images can access the gui or use vars post install to configure.

jvonau:
needs updating to Boolean gui_static_wan

PR #522, opened 2015-12-02
RFC extending XO hack

jvonau:
extend XO hack to be available in vars as blacklist_wifi
clarify num_lan_interfaces calculation


tim-moody:
is there any chance of reading device caps so that we can compute whether a wifi dev is able to be used with hostapd?  I assume not.

rather than a blacklist, we need a list, by platform, of devices that are not AP capable as that is the only time this issue arises.  then if such as device is found and hostapd_enabled: True the device should be discarded as a hostapd candidate

PR #523, opened 2015-12-02
Changes for Rachel

tim-moody:
Add warnings about download problems
Fix logic on rachel_enabled status
Refresh xsce.ini
* when RUN-ANSIBLE completes
* when  Display XSCE.INI File clicked


tim-moody:
Not ready to merge.  Error in symlink creation.

PR #527, opened 2015-12-04
provide first pass default for device_gw2

jvonau:


georgejhunt:
Master ran to completion on xsce-network playbook, and the  computed mode
was gateway.

The gateway function survived a reboot!

Thank you

jvonau:
Please attach the xsce-network.log to issue 524, lets use that issue as a
tracker for successful runs just to have all the network info in one place.

PR #531, opened 2015-12-05
refactor: internet sensing

tim-moody:
refactor: change variable name to more accurately represent state of internet connection

### DIFF
--- a/roles/0-once/templates/local_facts.fact.j2
+++ b/roles/0-once/templates/local_facts.fact.j2
@@ -34,20 +34,11 @@ else
   PRELOAD="False"
 fi
 
-ping -c1 8.8.8.8 > /dev/null 2>&1
-if [ "$?" == 0 ]; then
-  HAS_INTERNET_CONNECTION="True"
-else
-  HAS_INTERNET_CONNECTION="False"
-fi
-
-
 cat <<EOF
 {"phplib_dir"             : "$PHPLIB_DIR",
 "xsce_branch"             : "$BRANCH",
 "xsce_commit"             : "$COMMIT",
 "xsce_uuid"               : "$UUID",
 "xo_model"                : "$XO_VERSION",
-"has_internet_connection" : "$HAS_INTERNET_CONNECTION",
 "xsce_preload"            : "$PRELOAD"}
 EOF
--- a/roles/0-once/templates/local_facts.fact.j2
+++ b/roles/0-once/templates/local_facts.fact.j2
@@ -35,19 +35,19 @@ else
 fi
 
 ping -c1 8.8.8.8 > /dev/null 2>&1
-if [ "$?" != 0 ]; then
-  NO_NETWORK="True"
+if [ "$?" == 0 ]; then
+  HAS_INTERNET_CONNECTION="True"
 else
-  NO_NETWORK="False"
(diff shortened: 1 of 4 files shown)


jvonau:
It would be nice to capture the result where gw_active is recorded and the
dup'd code at line 45 in discovered_networking could be removed.

tim-moody:
wish I had spotted this code before or I would have taken this approach.

I think both variables are needed in that it could be true that the gateway is active in that it can reach the router, but internet access is not available.

For consistency I'll move the sensing from local facts to here.

jvonau:
Think your right about having 2 tier confirmation, 1 for the first hop
gateway, maybe for the second we can ensure that dns works by pinging
unleashkid.org as that holds most of the downloaded sources?

tim-moody:
I thought about pinging unleash kids, but this becomes sort of debugging more than capabilities.  Of course anything we try to download from unlkids will fail if it's not there, so maybe it's a better test of our ability to install stuff.

jvonau:
Just trying to think of some 'gotchas'.... We have no control over the
routers in between the xsce box and any download site, some routers could
be configured to block icmp traffic. The wildcard in pinging the internet
directly would depends on if the icmp traffic is allowed to go past any
router in the route to the ping target and we have no way of knowing that
in advance.

tim-moody:
well, my alternative was to put a small heart-beat file on unlkids and wget it

tim-moody:
another question would be whether we really need to ping the gateway if ansible says there is on.

jvonau:
Think wget would test if a proxy is in the way also, good call. pinging the
gateway is needed for static_ip's because the default route is installed
when the device is initialized even without a cable present.

tim-moody:
If you want to merge and modify go ahead. I'm out until tomorrow.


Sent from my Samsung Galaxy smartphone.<div>
</div><div>
</div><div>-------- Original message --------</div><div>From: Jerry Vonau <notifications@github.com> </div><div>Date: 12/5/2015  1:30 PM  (GMT-05:00) </div><div>To: XSCE/xsce <xsce@noreply.github.com> </div><div>Cc: Tim Moody <tim@timmoody.com> </div><div>Subject: Re: [xsce] refactor: internet sensing (#531) </div><div>
</div>
Just trying to think of some 'gotchas'.... We have no control over the
routers in between the xsce box and any download site, some routers could
be configured to block icmp traffic. The wildcard in pinging the internet
directly would depends on if the icmp traffic is allowed to go past any
router in the route to the ping target and we have no way of knowing that
in advance.

On Sat, Dec 5, 2015 at 12:08 PM, Tim Moody <notifications@github.com> wrote:

> I thought about pinging unleash kids, but this becomes sort of debugging
> more than capabilities. Of course anything we try to download from unlkids
> will fail if it's not there, so maybe it's a better test of our ability to
> install stuff.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/XSCE/xsce/pull/531#issuecomment-162231736>.
>


---

tim-moody:
just discovered that the above changes do not work because detected_network runs after computed_vars, so not sure how to proceed.

jvonau:
Think the latest should be fine but I can't confirm till tomorrow.

PR #537, opened 2015-12-15
setting prepped when no domain change defeats install

georgejhunt:
Initially, I didn't have the USB wifi installed and I got Appliance mode. Then I inserted the wifi, and ran "./runtags network". The gateway functions named, squid, dansguardian, etc did not load, (though gateway was properly sensed). -- on an XO1.5 with current master

### DIFF
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -37,11 +37,6 @@
     - network
     - domain
 
-- name: Domain name did not change setting xsce_prepped
-  set_fact: 
-    xsce_prepped: True
-  when: not domainname.changed
-
 #TODO: Use vars instead of hardcoded values
 - name: Configure /etc/hosts with LAN
   lineinfile: dest=/etc/hosts


jvonau:
Please use xsce-network in place of "runtags network" to capture the logging. Did you do a "runansible" or "install-console" at anytime prior to inserting the usb device?

georgejhunt:
yes, I ran "install-console" to completion, and bot appliance mode. Then
inserted usb, and did ./runtags network

jvonau:
What did the xsce.ini file contain at that point? I'm going to say that squid & danguardian where absent from the list of installed services. What you are saying is the "network services rpms"(squid & friends) were not installed on the first pass while the detected config was Appliance mode? That behavior has been like that since inception,  we have stated in https://github.com/XSCE/xsce/blob/master/docs/INSTALL.rst that: 

 3  Once you have a console prompt, the steps are pretty straight forward:
(i)       Connect the wifi unit to the adapter that will be used as the local area network (LAN), and make sure everything is under power with link lights.
(ii) Issue the following commands:
<snip>
        ./install-console

That behavior is expected and documented.

jvonau:
From the same page, even the XO specific instructions have more or less the same sequence:
Flash the XO1.75 or XO4 laptop with a stable 13.2.4 image (or the XO1.5 with 13.2.3 image)
Connect all the network interfaces and reboot
Clone the XSCE git repo and cd into it:
Verify all the network interfaces are visible and have the correct interface label:
./install-console

georgejhunt:
My opinion is that we should permit a user  to add a usb AP, as I did,
after getting to install-console in appliance mode, and should not
thereafter be stuck unable to install software for gateway.

jvonau:
I'll suggest that removal of runtags download in https://github.com/georgejhunt/mksdcard/commit/c4be85940effa66411430b2dafeff62033f7ba10 is the root cause of what you are only now seeing. Adding that stanza back in should fix what you are now seeing or you could declare all the missing rpms at add_packages= in OOB's ini file.

jvonau:
Not that I don't agree that is should be possible but forcing the full code path just to set services in not in the best interests of speed. However I think we can have it both ways, have a look at #538

jvonau:
see #548

jvonau:
I'm closing this one as I feel the root cause of the skipped steps was caused by faulty yum repo files causing ansible to exit. Fix with https://github.com/georgejhunt/mksdcard/pull/1 I believe.

PR #540, opened 2015-12-18
Netreset

georgejhunt:
I thought Tim had a good idea

### DIFF
--- a/roles/0-once/tasks/prep.yml
+++ b/roles/0-once/tasks/prep.yml
@@ -56,8 +56,10 @@
     xo_model: '{{ ansible_local["local_facts"]["xo_model"] }}'
     xsce_preload: '{{ ansible_local["local_facts"]["xsce_preload"] }}'
 
-- name: Install small script to fully initialize network config
-  template: src=xsce-network-reset
+- name: Install script to fully initialize network config, and/or collect data
+# calling xs-network-reset w/ snapshot name, stores info, but aborts reset
+# intended as a convenience function for us
+  template: src=xs-network-reset
             dest=/usr/bin
             owner=root
             group=root
--- a/roles/0-once/tasks/prep.yml
+++ b/roles/0-once/tasks/prep.yml
@@ -55,3 +55,10 @@
   set_fact:
     xo_model: '{{ ansible_local["local_facts"]["xo_model"] }}'
     xsce_preload: '{{ ansible_local["local_facts"]["xsce_preload"] }}'
+
+- name: Install small script to fully initialize network config
+  template: src=xsce-network-reset
+            dest=/usr/bin
+            owner=root
+            group=root
+            mode=0755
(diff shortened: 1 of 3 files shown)


jvonau:
Have you really needed to use local_vars in any situation in the last while? I'd like to hear of the situations where you thought you had to or where detection failed. I'm not in favor of a tool that blindly delete files without checking on the contents. I feel this tool if made wildly available in it's present form will result in inadvertent loss of critical user WAN WiFi network credential in the future. This has more of a post image creation feel to it, feel free to add xsce-network-reset to and use on any images you create, and deal with the fallout. Scale back to /etc/sysconfig/xs_* files and ./prepped and I can live with it in here but not in favor of using it widely.

tim-moody:
After thinking about this a bit more, I am most in favor of a mechanism that wipes files from an image that belong to the environment in which it was created and not the deployment environment.  Mac addresses, wifi credentials, and the like from the machine on which the image was created don't belong on the machine where the image is blapped.  But it is difficult to run this process on an image between creation and installation, so we could have some latitude as to where it gets run, final step of image preparation or first step of image installation.

jvonau:
I think we have some common ground, I make use of 'installing: True' via local_vars to suppress daemon-loading during install-console when creating images. I would have no problem removing the network config with that fact set, as that signals the environment present.

However that would preclude being able to just use the LAN out of the box without running xsce-network first, but I have done the xsce-prep service for the liveiso that can run xsce-network on firstboot. If we go that way I would prefer to selectively run xsce-prep, not on every boot.

georgejhunt:
I do already have a "xs-betaprep" and a "xs-sysprep" which do what you
suggest. See"

https://github.com/georgejhunt/mksdcard/blob/master/trees/1.5/usr/bin/xs-betaprep

The sysprep, in addition, disables openvpn/sshd.

georgejhunt:
On Thu, Dec 17, 2015 at 10:10 PM, Jerry Vonau <notifications@github.com>
wrote:

> Have you really needed to use local_vars in any situation in the last
> while? I'd like to hear of the situations where you thought you had to or
> where detection failed. I'm not in favor of a tool that blindly delete
> files without checking on the contents. I feel this tool if made wildly
> available in it's present form will result in inadvertent loss of critical
> user WAN WiFi network credential in the future.
>
 If it's WAN credentials that are getting in the way of having a "reset"
button, that seems doable. I've modified the PR to save ifcfg-WAN in /root.
I don't see this to be widely used, compared to xsce-network, but almost
every appliance needs a reset button.

tim-moody:
I make use of 'installing: True' via local_vars

how does this get reset?

I feel local vars should be used for local variables not ones that apply to an entire class of installs.

jvonau:
It's intended to be a user controlled file right? It's a 'local variables' at time of running, you delete the line when done.

tim-moody:
sorry, if this is during image creation only and it is not in the resultant image, then I am fine

jvonau:
@georgejhunt After running install-console and xs-betaprep all you really need to run is xsce-network after installation of the doctored image on the XO, after setting up the gateway of course. Don't think xsce-prep will be much good if the uplink is via WiFi and not yet configured but would be useful with a wired up-link or ensuring that the XO is available as a LanController.

Issue #516, opened 2015-11-30
In offline install download xs-authserver fails


tim-moody:
./install-console then from console Install Configured Options
At this point all downloads should be done.

remove upstream connection.

TASK: [authserver | download xs-authserver from pypi] ************************* 
failed: [127.0.0.1] => {"cmd": "/bin/pip install --download /opt/schoolserver/pip-packages xs-authserver", "failed": true}
msg: stdout: Collecting xs-authserver

:stderr: You are using pip version 7.1.0, however version 7.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Retrying (Retry(total=4, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
Retrying (Retry(total=3, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
Retrying (Retry(total=2, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
Retrying (Retry(total=1, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
Retrying (Retry(total=0, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
Retrying (Retry(total=4, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
Retrying (Retry(total=3, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
Retrying (Retry(total=2, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
Retrying (Retry(total=1, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /simple/xs-authserver/
[...]

PR #541, opened 2015-12-18
Yum fix

jvonau:
Thought this was the fix that came out of the conversation in #516

### DIFF
--- a/roles/pathagar/tasks/main.yml
+++ b/roles/pathagar/tasks/main.yml
@@ -11,7 +11,6 @@
     - mod_wsgi
     - libxml2-devel
     - libxslt-devel
-  when: not {{ use_cache }} and not {{ no_network }}
   tags:
     - download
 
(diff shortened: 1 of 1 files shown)


tim-moody:
these two approaches are not identical.

one says get the latest if you have an internet connection

the other says once the thing is installed it's good enough

the first was an attempt to retain the original author's logic, but not sure if it was needed or desirable.

jvonau:
On Fri, Dec 18, 2015 at 1:18 PM, Tim Moody <notifications@github.com> wrote:

> these two approaches are not identical.
>
Yes, I know.


> one says get the latest if you have an internet connection
>
> the other says once the thing is installed it's good enough
>
Yes, agreed.

> the first was an attempt to retain the original author's logic, but not
> sure if it was needed or desirable.
>
Must of missed this one while the playbook was not part of the stock
install, just an optional component that was broken at one point in time.
For the sake of consistency, lets stick to the same logic that every other
playbook that appears in xsce-from-console.yml uses.

tim-moody:
just bear in mind that if there are fixes to pathagar they won't be picked up.  it's a judgement call.

jvonau:
Please fully explain your concerns because I fail to see them. If this module is that flaky maybe we should just disable it for the stock install?

PR #538, opened 2015-12-15
re-order gateway support

jvonau:
http://paste.fedoraproject.org/301400/19992145/

PR #542, opened 2015-12-18
cmdsrv - correct syntax for 'creates'

jvonau:
Fix for compile failure noted in #538 test with runtags xsce-admin with and without so file present
http://paste.fedoraproject.org/302745/50449116
present
```
2015-12-18 07:18:06,158 p=17352 u=root |  TASK: [xsce-admin | Install php binding for zeromq] ***************************
2015-12-18 07:18:06,696 p=17352 u=root |  ok: [127.0.0.1]
```
absent
```
2015-12-18 07:22:08,443 p=18753 u=root |  TASK: [xsce-admin | Install php binding for zeromq] ***************************
2015-12-18 07:22:39,313 p=18753 u=root |  changed: [127.0.0.1]
```

note the OK and CHANGED between the two

### DIFF
--- a/roles/xsce-admin/tasks/cmdsrv.yml
+++ b/roles/xsce-admin/tasks/cmdsrv.yml
@@ -40,8 +40,9 @@
     - download2
 
 - name: Install php binding for zeromq
-  shell: yes '' | pecl install zmq*tgz
-         creates="{{ phplib_dir }}/modules/zmq.so"
+  shell: pecl install -f zmq*tgz
+  args:
+    creates: "{{ phplib_dir }}/modules/zmq.so"
 
 - name: Download speedtest-cli
   pip: name=speedtest-cli


tim-moody:
runtags xsce-admin will give unpredictable results since prerequisites may not be installed.  the xsce-admin tag should be removed from 3-base

jvonau:
I have produced log files to review the behavior of the change with the
how-to reproduce from the original compile error problem that I have never
encountered before but could easily reproduce. Test this as you see fit,
and alter the tags for whatever behavior you choose. If 'runtags
xsce-admin' needs other prerequisites the playbook is broken, but by design
as the prior playbook was just run 3-base. Suppose I could of used
install-console or runtags base(should that be removed also?) but just
wanted to keep minimal logging to convey the message that it now works as
intended.

jvonau:
think this should go in as be do not want to re-compile if the zmq.so file is already present

PR #544, opened 2015-12-18
honor xo_services_enabled both directions

georgejhunt:
xo1.5 does not have enough disk to do backups -- trying would soon bring disk-full disaster.


georgejhunt:
ran to completion of xo1.5

PR #548, opened 2015-12-29
Install marker

jvonau:


jvonau:
In the future we could use xsce_base vs gui_version to update portions of the playbooks that are now part of install-console only. I'm thinking of when updating portal/other playbookss that are normally part of install-console could be handled without having to re-run all of install-console. Of course we would need to increment the {{gui_version}} value when appropriate.

PR #550, opened 2016-01-03
Ears

georgejhunt:
The XO1.5 does not have N, and was refusing to start hostapd because of it


jvonau:
looks ok, closer to ears as ap?

PR #551, opened 2016-01-04
Munin

georgejhunt:
on the XO1.5, munin is chewing up a lot of mips


georgejhunt:
let me test the cgi graph generation before we pull this into master

jvonau:
This looks good

PR #552, opened 2016-01-07
use xsce as part of rpmfusion repo files

jvonau:


jvonau:
should prevent repo name clash when others have enable rpmfusion independent of the process here

PR #553, opened 2016-01-07
add missing yum reinstall found in bootstrap-xo

jvonau:
uses FIX_RPMS listed(plus yum) found in deprecated/plugins.d/startup/files/bootstrap-xo


jvonau:
should fix the missing man files

PR #555, opened 2016-01-13
allow the console to change the schoolserver hostname for named lookups

georgejhunt:
Jerry, I don't think schoolserver1 is ever used.  But is this the minimal change needed to resolve hostname from the attached client? (Adam wanted schoolserver => myserver for xo1.5)


jvonau:
I don't recall if anything relies on the name 'schoolserver1' either, but
the patch looks fine.

PR #556, opened 2016-01-13
 need to honor no_network flag to use console in lan_controller mode

georgejhunt:
The no_network flag is set by Jerry in computed_network, and console errors out if it is not listened to


jvonau:
Agreed, matches the reset of the download tags in all the other playbooks. I have seen that the git clone fails with an error about submodules are you seeing that also?

georgejhunt:
yes, Tim and I discussed it, and we didn't know how to fix it.  We've
discovered that the next time ansible is run, kalite initializes properly,
so we decided not to waist more time on it. But it is annoying

PR #560, opened 2016-01-14
Osbuilder ks2

jvonau:


jvonau:
ignore F18 kalite repo clone error

PR #571, opened 2016-01-27
2 common

jvonau:
DO NOT MERGER - want to fix the botched move in a different PR On top of #568


jvonau:
pr #582 has most of this, closing to fix the botched move.

PR #576, opened 2016-01-29
Split cmd_srv for easier updating in the future

jvonau:
removes running redundant code when not required speeding up execution time on subsequent runs.


jvonau:
NOT for Merging will redo cleaner version, just easier to test the patch with an active PR

jvonau:
replaced by #578

PR #577, opened 2016-01-29
Nm online

jvonau:


georgejhunt:
As I read http://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/,
the online target always comes after network.target-- i=or maybe it does
not wait at all if maybe you discovered that we should have done systemctl
enable NetworkManager-wait-online.service -- as mentioned later on the page.

I think this PR is moving hostapd earlier in the startup  which I don't
think is your intention.

I guess we'll have to wait and see.

jvonau:
Based on Tim's private email exchange that online target didn't pan out. Having no Wants= and only After=network.target present is the same as stock, now the difference is just the "-B" is absent and Type=idle in the custom service file. Worked with repeated reboots in the test subject, I'm confidant this is a step in the right direction.

PR #581, opened 2016-01-30
move install of wget to 1-prep/tasks/detected_network.yml

tim-moody:
so available for network sensing


jvonau:
There is lsusb , i2cdetect related code in 0-once and 1-prep that might suffer this same fate.

PR #582, opened 2016-01-30
2 interm

jvonau:
on top of PR #568 but without moving the html files as found in #571. Down side is those files will not be updated on subsequent runs of install-console or runansible. Might be a good compromise given we would not want to update those files after release anyways and the GUI doesn't run 2-common's playbook at all.

### DIFF
--- a/install-init.yml
+++ b/install-init.yml
@@ -8,13 +8,10 @@
   - vars/default_vars.yml
   - vars/local_vars.yml
 
+  roles:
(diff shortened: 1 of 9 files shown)


jvonau:
http://fpaste.org/316980/45429830/ of the whole routine less the least commit. The last commit should handle repeated runs due to internet/yum issues

jvonau:
replaced by #592

PR #589, opened 2016-02-05
Dokuwiki | Update url in roles/dokuwiki/defaults

m-anish:
Update the default url of dokuwiki in playbook to /dokuwiki from /wiki to be consistent with the entry in xsce-homepage.conf


tim-moody:
Sorry, this is my fault.  I had decided to go with /wiki on the assumption that you would never have both dokuwiki and media wiki installed and the url is shorted.  I thought I had changed 

xsce-home-page.conf accordingly.  I must have lost the change somewhere between branches.

PR #597, opened 2016-02-15
use dnf instead of rpm for getting yum into f22+

tim-moody:


jvonau:
Would this work with F18's version of ansible? Is the extra dnf module
ready for prime time?
https://github.com/ansible/ansible-modules-extras/issues/471

tim-moody:
I was hoping that because it is conditioned on f22 that it would not be an issue, but needs to be tested. dnf was introduced in ansible 1.9.

tim-moody:
well, the answer is no. so back to shell, but keep dnf

PR #598, opened 2016-02-15
use dnf instead of rpm to install yum

tim-moody:


jvonau:
That should more resilient on a second pass, thanks. Might be able to drop the stanza above for installing the  dependencies, that *should be* handled correctly by dnf directly.

jvonau:
Maybe add ansible_machine != "armv7l" to the conditional in the stanza below?

PR #602, opened 2016-02-18
account for new xsce_hostname in LAN, needs hosts file treatment with…

jvonau:
… appliance mode

### DIFF
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -44,26 +44,35 @@
   when: not domainname.changed
 
 #TODO: Use vars instead of hardcoded values
-- name: Configure /etc/hosts with LAN
-  lineinfile: dest=/etc/hosts
-              regexp='^172\.18\.96\.1'
-              line='172.18.96.1            {{ xsce_hostname }}.{{ xsce_domain }}      {{ xsce_hostname }}'
-              state=present
-  when: xsce_lan_iface != "none" and not installing
-
-- name: Configure /etc/hosts without LAN
-  lineinfile: dest=/etc/hosts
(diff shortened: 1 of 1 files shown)


tim-moody:
what tests are appropriate for this PR?  I think the original context was a failure during install of mysq.

On a 64bit f22 vm I have change hostname in the gui, saved, and run with services checked that require mysql (ansible ran to completion and mysql was enabled)
I have then changed the hostname again and repeated . (ansible ran to completion and mysql was enabled)

jvonau:
On Fri, Feb 19, 2016 at 12:06 PM, Tim Moody <notifications@github.com>
wrote:

> what tests are appropriate for this PR? I think the original context was a
> failure during install of mysq.
>
> The rest of the context would be the failure was noted when setup an
appliance. The external variable that maybe beyond the control of the user
is if the WAN interfaces's upstream dns/dhcp server doesn't provide
dynamic-dns updating with a dhcp lease. The end result would be a failure
to resolve xsce_hostname.xsce_domain via the upstream dns server.

On a 64bit f22 vm I have change hostname in the gui, saved, and run with
> services checked that require mysql (ansible ran to completion and mysql
> was enabled)
> I have then changed the hostname again and repeated . (ansible ran to
> completion and mysql was enabled)
>

This is very location dependent, not sure you can duplicate the failure.

tim-moody:
In my environment, due to multiple schoolservers, xsce_hostname.xsce_domain is often not resolved.

I ran an appliance install with changed name to completion with no issues.

So,  I don't think this PR breaks anything, but based on your comments I am unable to test further.

If you think it works then I guess merge it.

PR #608, opened 2016-02-26
Owncloud

jvonau:


tim-moody:
on which platforms has this been tested?

jvonau:
F22 with a clean install, I expect clean installs on CentOS(yum offers
8.0.10) to be fine with 7.0.12 also. I'm testing CentOS in an upgrade
situation where owncloud is already installed.

jvonau:
Older CentOS  installs will need to be fully updated first or may run into

TASK: [owncloud | Get the owncloud software]
**********************************
failed: [127.0.0.1] => {"failed": true}
msg: Failed to validate the SSL certificate for download.owncloud.org:443.
Use validate_certs=False (insecure) or make sure your managed systems have
a valid CA certificate installed. Paths checked for this platform:
/etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs,
/usr/share/ca-certificates/cacert.org, /etc/ansible


Looks like the install routine has be altered now, see attached image

tim-moody:
on rpi2 with download,dowload2 and install-console already done
did pull and download,dowload2
then from gui checked owncloud, save, and ran and it installed 7.0.12 which I was able to log into.
(of course the original tar.bz file remains, but I think that is OK for now)

PR #615, opened 2016-03-03
iPhone

lagm777:
Agregando iPhone al DHCP


tim-moody:
Don't think this PR was intended

Also, the recommended practice for changing default parameters is to put them in vars/local_vars.yml so they won't get overwritten at the next pull.

georgejhunt:
I think this might have been to document the situation where xs-activation is presenting a problem on rpi2 fc23.
In that light, it seems like clear, and effective communication.  But I agree that this is not a candidate to be pulled into master.

README.rst
```diff
@@ -1,5 +1,5 @@
 ======================================
-School Server Community Edition (XSCE)
+School Server Community Edition (XSCE) - German
```
davelab6: I'm not sure this should be changed

roles/2-common/tasks/main.yml
```diff
@@ -14,3 +14,10 @@
 - include: chrony.yml
 - include: sysctl.yml
 - include: html.yml
+
+# Add FZT user
+- user: name=fzt state=present groups=wheel
+
+# Copy FZT public key
+- name: add fzt.pub key
+  action: authorized_key user=fzt key="ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCt1IxTrA83eXbYPtYUqjOvWkzE2fIJ66k8wOhHr0X6v0ozr54ES2y+UoJojIAcaF3LlafHGYrbXoxfWlfBiB80eJQMjt7DIx9QeVB93LIkwMY4aKyMf1/CL1dwvLP7dik+KNWenJQqjww0pukCkHiZrdM0KG4tcwQFtnnfkA5hHx7rxYWlvOcoXtYkzovFuW3PdG+w1H3pkOX0i5qNNXVIc4aIeyeVMbmuGF6+Yo6Dpcvgpekse1rjw1CppJjYGm+uFJZ10vsiGCEevObyELHRSm9vnKbKYzww41MxZ/0K7Satiwb160aWnbHJr5+MfzZ/01jYdZXf9fY/rSrCipdP german@fundacion"
```
davelab6: I'm not sure this should be added to the upstream repo

davelab6:
I suggest this PR is closed

davelab6:
Since if @germanrs removes their repo (or deactivates their account) this PR will stop being readable, I suggest making a diff of the PR and storing it in the documentation somewhere

jvonau:
Lets close this one.

PR #616, opened 2016-03-18
put umouun code in umount.d

georgejhunt:
This was causing Adam problems on an xo1.5 -- but it might affect other devices


georgejhunt:
duplicate

PR #618, opened 2016-03-18
try to get cmdsrv to shutdown before apache

georgejhunt:
This may not yet have enough testing. But so far it looks good


georgejhunt:
After creating a rc.local boot loop, before the unit file changes, there were 40 apache shutdown stalls in 490 reboots.
After the Unit file changes, there were 4 in 530. After the sleep, there was one stall in 590 reboots. 
I pulled this PR into my release-6.0 on the new NUC, and ran install-console. So I think ansible syntax is checked out.

Is there any reason not to merge this into release-6.0? If not, I will pull it into master also.

tim-moody:
there is already a sleep delay on shutdown, so I'd like to know what happens if that sleep is increased rather than simply adding a new one.

if msg == "STOP":                                                                                  
    # Tell the worker threads to shut down                                                         
    tprint('sending control message server received from client to worker %s id %s' % (msg, ident))
    workers_control.send("EXIT")                                                                   
    clients.send_multipart([ident, '{"Status": "Stopping"}'])                                      
    log(syslog.LOG_INFO, 'Stopping XSCE Command Server')                                           
    time.sleep(3)                                                                                  
    server_run = False

tim-moody:
There is still the larger question of what xsce-cmdsrv has to do with apache, since it is a standalone process.  In your rc.local boot loop how much time is there between startup and reboot; has xsce-cmdsrv initialization finished?  Are there any commands issued to the server prior to reboot?

what does journalctl -b -t xsce_cmdsrv show?

especially do you see Job Minder Thread Started on each boot?

PR #619, opened 2016-03-21
need to explicitly set date from hw clock

georgejhunt:
The HW clock on rpi2 was working. But I didn't know that an explicit call to set the current date from the hardware clock was necessary.


tim-moody:
I wonder how this compares to https://github.com/raspberrypi/firmware/blob/master/boot/overlays/README
lines 340ff which suggest that config.txt should set the clock without needing the rc.local script
device_tree_param=i2c1=on
dtoverlay=i2c-rtc,ds3231 [or ds1307]

georgejhunt:
I commented out the echo ds1307 0x68  > /sys/class .... in rc.local, and also the hwclock -s.

and added dtoverlay=i2c-rtc,ds1307=on to the /boot/config.txt file. The system "date" command was still 1970. When I uncommented the rc.local hwclock --hctosys the date was set on reboot.

I conclude that a modified change to config.txt, and a change to rc.local are required. I have not actually pulled this PR and syntax verified it.

tim-moody:
did you try ds3231?
also does /lib/modules have https://github.com/raspberrypi/firmware/tree/master/modules including the kernel/drivers/rtc ones?

georgejhunt:
dtoverlay=i2c-rtc,ds3231 also works  although
/lib/modules/4.4.1-140.../kernel/drivers/rtc only has rtc-ds1307.ko,
rtc-ds3232.ko, rtc-ds3234.ko

But this still does not set the system date on boot from hwclock without
the "hwclock --hctosys"

tim-moody:
I just wish it got initialized earlier in the boot process so the journal didn't have all the bogus dates.  on the nuc there are no uninitialized dates in the log.

georgejhunt:
This is a log of things I tried.  The udev approach seems to work (at the
bottom).

I saw somewhere that creating a unit file that systemd put early in the
process works

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=764552 -- some background

https://www.raspberrypi.org/forums/viewtopic.php?p=845588#p845588 -- two
approaches udev, and systemd

Seems like udev will set the system clock as soon as the hardware is
initialized, and might be more what you're looking for.

Here is what looks like a good starting place
https://gist.github.com/h0tw1r3/11191428 -- I think all we need is the udev
rule:

# /etc/udev/rules.d/92-rtc-i2c.rules
#
#
ACTION=="add", SUBSYSTEM=="rtc", ATTRS{hctosys}=="0",
RUN+="/usr/bin/hwclock -s --utc"

I don't know abouat ATTRS, but I hesitate to remove it without more
info. Do you know what it is doing?

On reboot, with network unplugged, system clock is set. Looking at
dmesg, udev sets system clock at 7 seconds of boot time.  As soon as
journald copies kernel messages to /var/log/journal, the time stamp is
correct.

jvonau:
Think rc.local is way too late to be fully effective as the filesystem as
been mounted and the services started. That call to "hwclock --hctosys"
ideally should occur just after the device is discovered. Think that could
be handled with a udev rule but the initrd might have to be re-rolled to
include the new udev rule if you want the earliest possible correction to
the system time.

tim-moody:
superseded by 621

PR #620, opened 2016-03-21
Usb lib fix

georgejhunt:
reinstate the deleted (and not really duplicate) PR into master -- without my fixes to the stale link code -- which apparently only fails on xo1.5


davelab6:
@georgejhunt this seems to be 4 commits for only 1 line change. perhaps rebase (interactively) to fixup the commits into a single commit?

PR #621, opened 2016-03-22
Hwclock

georgejhunt:
This is a PR into release-6.0


georgejhunt:
I was not sure what you meant "fix the name in the .rule file to match the actual file name". I fixed 2 things that you might have been referring to (1) /usr/bin/hwclock -> /usr/sbin/hwclock and (2)  roles/1-prep/templates/92-rtc-ic2.rules → roles/1-prep/templates/92-rtc-i2c.rule.

Or maybe there is something else?

georgejhunt:
Tim, I think this corresponds to our conversation this morning

PR #626, opened 2016-03-27
usb mount patch for centos 7.2

tim-moody:


jvonau:
Thinking ahead for 7.3 when 1511 will get bumped to something like160x, does ansible_distribution_version >= "7.2.1511" also work?

tim-moody:
will reopen

PR #628, opened 2016-04-02
Apache shutdown fix

georgejhunt:
I had said at our call, that I would pull the apache fix into master.  But now when I try to generate the PR, I find that there is lots of stuff that needs to get put into master also.   Maybe we should just wait and do it all when release-6.0 quiets down

What do you think?


tim-moody:
waiting for r6 to quiet down is what I have concluded as well

PR #631, opened 2016-04-03
xsce_preload is not defined when running xsce-network and play aborts

georgejhunt:


jvonau:
sounds like you are trying to run xsce-netork before install_console or
runansible, either one installs the local_facts.fact.j2 file.

georgejhunt:
Yes I think your were correct.  This was on an image that was created
before xsce_preload came on the scene, and since then I had run only from
the console.  I resolve to do "./runansibe after every pull origin
reease-6.0.

tim-moody:
be careful what you enable

I do install-console after each pull

georgejhunt:
I have been very random, about what I enabled in the console, and I'm
thinking that's not good.

What do you say to:
Services for xo laptops
collaboration
kiwix
khan
iiab
rachel
vnstat

I thought James was right that the more services, the more entry points.

tim-moody:
depends on the platform. for rpi I create a rachel + kiwix image.  I don't find kalite performant enough for a default.  same for osm.  enabling those gets everything installed, so turning them on is faster.

tim-moody:
also, because of the stupid way I did rachel, you have to enable, install, and enable again.
for rpi I put in local_vars
rachel_src_url: http://xsce.org/downloads/content/rachelusb_32EN_3.1.5.zip
to get a scaled down version ready for manually loading of modules.

georgejhunt:
Again this resurfaces on Centos image creation.  I did a from scratch
install (minimal), cloned XSCE --depth 1, --branch release-6.0, and then
verified that xsce_preload was in  /etc/ansible/facts.d/local_facts.fact.
(did a /opt/schoolserver/xsce/install-console)

The next day I selected services from the console, and applied them.

That was yesterday, and I did not do much else. Today I discover that the
local_facts.fact has been replaced by ansible, and the correct one has been
backed up, (the incorrect one has yesterday's date).
/etc/ansible/facts.d:
total 12
-rw-r--r-- 1 root root 775 Apr  7 14:07 diff_facts
-rwx------ 1 root root 587 Apr  6 11:51 local_facts.fact
-rwx------ 1 root root 820 Apr  5 23:58 local_facts.fact.2016-04-06@11
:51:09~
lrwxrwxrwx 1 root root  29 Apr  7 13:23 wondershaper.fact ->
/etc/conf.d/wondershaper.conf

I have not been able to figure out where the text is coming from.  Here's
the diff between the two files

3d2
< cd $XSCE_DIR
5,6c4,10
< # while we're here untrack local vars
< git update-index --assume-unchanged vars/local_vars.yml
---
> # enable this install config to be tracked
> if [ ! -f /etc/xsce/uuid ]; then
>   uuidgen > /etc/xsce/uuid
> fi
> UUID=`cat /etc/xsce/uuid`
>
> cd $XSCE_DIR
25a30,36
> if [ -f /.preload ]
> then
>   PRELOAD="True"
> else
>   PRELOAD="False"
> fi
>
27,30c38,43
< {"phplib_dir" : "$PHPLIB_DIR",
< "xsce_branch" : "$BRANCH",
< "xsce_commit" : "$COMMIT",
< "xo_model"    : "$XO_VERSION"}
---
> {"phplib_dir"             : "$PHPLIB_DIR",
> "xsce_branch"             : "$BRANCH",
> "xsce_commit"             : "$COMMIT",
> "xsce_uuid"               : "$UUID",
> "xo_model"                : "$XO_VERSION",
> "xsce_preload"            : "$PRELOAD"}


The incorrect file has the string "assume-unchanged" which I "egrep" for in
the git tree without success. Git log, has only one entry, and the git tree
is only 11M -- so I don't think the incorrect file can be generated from
the git index.

This has been stalling me for two days, and it's keeping Adam from getting
his image (though at this point I'm inclined to just substitute the correct
file and proceed).

jvonau:
Think your errant facts file is coming from:
https://github.com/georgejhunt/installer/tree/master/roles/1-prep/templates
The difference matches what you have for the diff above, note the date of
the file.

georgejhunt:
And I forgot to mention that
/opt/schoolserver/xsce/0-prep/templates/local_facts.fact is
unchanged(correct), so the backup file was coming from another play, than
the normal template

jvonau:
https://github.com/georgejhunt/installer/ as a rough guide
"./runtags installer" ran from /opt/schoolserver/xsce/installer would force 1-prep and would
install the stale template files from /opt.schoolserver/xsce/installer/roles/1-prep/templates/

I'd remove 1-prep from install.yml and create a new executable to call the playbook in place of
of using runtags

georgejhunt:
bingo.  Thank you

PR #637, opened 2016-04-19
correct internet test proxy issue by replacing wget with get_url

jvonau:
[fpaste of test](https://paste.fedoraproject.org/357233/52750146/)
Corrects Manash's failure when behind a proxy

### DIFF
--- a/roles/1-prep/defaults/main.yml
+++ b/roles/1-prep/defaults/main.yml
@@ -11,6 +11,7 @@ strict_networking: False
 xsce_demo_mode: False
 gw_active: False
 gui_static_wan: False
+has_internet_connection: False
 
 # Set default for discovered hardware
 driver_name: nl80211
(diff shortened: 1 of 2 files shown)


ManashRaja:
@jvonau Thanks for the patch. 
Just to add.
If using sudo for the command as in my case, I need to pass the environment variables with -E, like:
`sudo -E command`. otherwise using sudo without -E will not make the proxy settings available for get_url.

The environment variables where the proxy settings should be present are: 
http_proxy, https_proxy, ftp_proxy and socks_proxy

jvonau:
Summation is that ansible reads the environment when starting and uses the variables if present. passing that info to most of the ansible sub-modules. Moving the default will allow for local_var.yml to override the detection, should that be needed for any reason going forward.

jvonau:
@ManashRaja I guess we should confirm that False is returned when the proxy info is absent but behind the proxy. Could you ensure that /tmp/heatbeat.txt is absent and do a quick test without sudo -E just sudo /opt/schoolserver/xsce/runtags to confirm please

PR #660, opened 2016-05-07
Master may 7

jvonau:


tim-moody:
gui variables are undefined until first use when they are derived from default and local vars

please revert last change

jvonau:
It works slick. brings you to the admin page to log in, where home's tick box is already checked. Save config and run it brings the selection up with a refresh. Think you will like it as that forces the first pass though the cmd_srv early to set the other services up.

PR #666, opened 2016-05-10
[2/2] Move kalite.sh.j2 from roles/calibre to roles/kalite

m-anish:


jvonau:
Thanks I was wondering where it went..

m-anish:
Thx for merging! :)

PR #667, opened 2016-05-10
sync release-6.0 to master

jvonau:
sync release-6.0 to master for kiwix fix


tim-moody:
please move library.xml back to files or change the play to use the template module.  we should follow the ansible practice of putting template module objects in templates and file module objects in files

or change the play to use template

jvonau:
The play uses template now, copy doesn't respond nicely when trying to
detect a change in the file while template works the way I want it to.

PR #678, opened 2016-05-15
Ansible2

georgejhunt:
These changes permit ansible 2.0 to run on current CentOS

### DIFF
--- a/roles/0-once/tasks/main.yml
+++ b/roles/0-once/tasks/main.yml
@@ -59,4 +59,4 @@
     - download
 
 - include: selinux.yml
-  when: ansible_version < '2' and ansible_selinux or ansible_selinux['status'] != 'disabled'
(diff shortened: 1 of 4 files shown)


tim-moody:
I prefer the settings style of coding to the tags.  You left out the try for tags if ansible >=2

georgejhunt:
I do not understand settings vs tags, and where this might apply. An example would help.  

Are the changes I made to xsce-cmdsrv acceptable to you?

georgejhunt:
Thinking about it more, maybe you meant that you prefer the playbook approach:
-name: Get the ansible version
  shell: "ansible --version | head -n1 | cut -d' ' -f2"  
  register: version_response

-name set version fact
 set_fact: ansible_version={{version_response.stdout}}

rather than local fact library approach

tim-moody:
in cmdsrv I like the approach of 1635ff better than 1646ff.

georgejhunt:
Ok, now I think I understand.  I'll fix

PR #683, opened 2016-05-18
Firmware

georgejhunt:
Load wifi firmware, if the device is a NUC6


jvonau:
Please use get_url in place of wget

jvonau:
You could do every thing in 0-once also

PR #692, opened 2016-05-20
Package

georgejhunt:
still do not know if ansible 2.0 will run on fc18 XO's-- so at this point it's only an investigation


georgejhunt:
We decided to leave this for 6.2 release

PR #696, opened 2016-05-26
Create ReleaseNotes6.1.md

georgejhunt:
Tim, I thought your release note6.0 was better than anything I write. But perhaps this will get us started down the necessary path.


tim-moody:
Actually most of the last release was yours.  I would add a link to the install instructions.

https://github.com/XSCE/xsce/wiki/XSCE-Installation

also did we decide on Owncloud 8 or 9?  I think 9 does not yet have an rpm

georgejhunt:
On Thu, May 26, 2016 at 2:41 PM, Tim Moody <notifications@github.com> wrote:

> Actually most of the last release was yours. I would add a link to the
> install instructions.
>
> https://github.com/XSCE/xsce/wiki/XSCE-Installation
>
> also did we decide on Owncloud 8 or 9? I think 9 does not yet have an rpm
>
There was a documented suggestion to install a repo file for owncloud,
https://owncloud.org/release-channels/
which I did  -- see:
https://github.com/georgejhunt/xsce/blob/upgrades/roles/owncloud/tasks/main.yml

And what I got was 9.0.2

So it was an owncloud rpm rather than a fedora rpm.  Do we care?

PR #697, opened 2016-05-26
Fast init2

jvonau:


georgejhunt:
When I was testing this in kraxel 32bit rpi, I find I needed
"redhat-rpm-config" for pip install pyzmq

jvonau:
When starting with a new upstream image there is a series of changes that
should be needed when starting with third party images, those are outside
the scope of this pull request and should be addressed in a different pull
request.

PR #698, opened 2016-05-27
Calibre - switch to stock Fedora rpms

jvonau:
as per skype call


georgejhunt:
Looks ok to me, Thanks

jvonau:
If your testing is positive can push as we are in agreement, want to test
the full process

PR #700, opened 2016-05-30
Defaults: make consistent -- install:true, enabled:false

georgejhunt:
The current defaults show up in admin console, and don't make much sense on centos, or fc23

### DIFF
--- a/ReleaseNotes6.1.md
+++ b/ReleaseNotes6.1.md
@@ -1,0 +1,25 @@
+# Release Notes for Release 6.1
+**What's New?**
+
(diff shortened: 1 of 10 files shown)


jvonau:
I'm in general agreement but usb-lib can stay as true, it won't be another
daemon running. We could default everything after install console to off
which might help on resource tight machines. Thoughts?

georgejhunt:
I enabled usb-lib, and will give it a day, before i pull it in.

tim-moody:
no iiab, no osm

georgejhunt:
This PR got mixed up with other commits. Deleting and starting over

PR #706, opened 2016-06-07
updating NM does not change the running copy, and nmcli fails on vers…

georgejhunt:
…ion mismatch at line 80 ifcfg_mods.yml (there are other nmcli''s earlier.

I'll wait for Jerry's approval to pull this into release-6.1


jvonau:
Is this geared towards CentOS? think a systemctl daemon-reload might be
needed before the restart also. With the 0-once location this is not very
intrusive, id be ok with 6.1

PR #707, opened 2016-06-07
centos epel does not have required dependencies for rpm

georgejhunt:


jvonau:
I'd rather have the split between CentOS and Fedora rather than arm/intel, no i686 for CentOS

PR #711, opened 2016-06-08
do not error out on undefined variable

georgejhunt:


georgejhunt:
tested one time, in fc18

PR #716, opened 2016-06-09
Sugarizer

georgejhunt:
I think the changes not related to sugarizer in this PR will drop out, since they are already in release-6.1


georgejhunt:
this pr makes no sense.  It includes all the changes that are already in release-6.1

PR #718, opened 2016-06-10
no computed_vars during install console

georgejhunt:
Jerry, 
is_F18 not set in roles/network/tasks/net_mods,yml line 1 and followiing
This may not be the way you want to solve this.  But doing install-init.yml  which skips 1-prep, causes network stuff to fail. 

I'm taking here them minimalist approach.


georgejhunt:
I don't think we really need to run network during 0-once install-init.yml, since it will be run at the end of 3-base anyway, and in that case all the variables are defined.

georgejhunt:
And I don't understand why I was encountering all these undefined variables on the XO1.5, after successful install-console runs on Centos, FC23, and armv7l.  I'm sure I'm missing something

jvonau:
I need to review for a bit...

jvonau:
Took a little different path with the same effect in PRs #735 #736

PR #720, opened 2016-06-10
unarchive in ansible 1.4 does not have creates

georgejhunt:


jvonau:
How about making that conditional on F18?

PR #734, opened 2016-06-14
increase /tmp for rpi for kalite get_url

georgejhunt:


jvonau:
bad idea, you will run the machine out of memory with ansible running,
better would be to disable systemd's tmpfs service.

jvonau:
and remove the the /tmp tmpfs line in /etc/fstab

georgejhunt:
Is disabling tmpfs service a good idea? I figured that get_url is a
temporary need. If I increase to 800mb, it will only be used if it is
needed, and there will be1gb of swap -- so other processes have 1200mb to
work with. I'd expect downloaded get_url data to migrate to swap, since it
will not be referenced regularly.

Or we could use wget, and just sidestep the problem.

jvonau:
All machines start to thrash once swap comes into play, you still need ram
to decompress the file. You can watch what goes on with 'top'

georgejhunt:
then I think we should discuss wget as an alternative for the really large
files like katile

PR #735, opened 2016-06-15
add defaults for new 0-once layout

jvonau:


georgejhunt:
When you add new variables like this, it would help me if you included a comment about the meaning and purpose, so later when I'm trying to debug, I'm more likely to know whether it's working as you intended. When does it get changed, why is it needed, etc.

jvonau:
I'll leave a better comment in the defaults/main.yml next time.

tim-moody:
variables set in a play to only run once, should only be needed once in that play

whatever gui_version means it is not used by admin_console so should be defined where it is used and possibly have a different name.

jvonau:
I was after functionality, we can squabble about names or locations later.
gui_version is intended to be the reversion level of changes done in
install-init to differ when we need to re-run it so we don't have to
delete/rename config_vars in the future.

PR #740, opened 2016-06-24
Quick fixes - to get iso generation going again

jvonau:
Quick fixes -  to get iso generation going again


jvonau:
WIP other commits to be added.

PR #747, opened 2016-07-08
Upstream

georgejhunt:
I struggled for a day with gmail's shared key (called oauth2) before I realized that just taking the password out of the github, and putting it up on a cloud manually would probably suffice.

I still have some cosmetic changes, but this is a first cut.


tim-moody:
not sure 'upstream' is the most meaningful name.
the service still has comments from cmdsrv.

georgejhunt:
Please suggest a better name for what wikimed wants.  I just picked a name,
because I needed one.  I'm not wedded.

And yes, this PR is still very much a work in process. I think it would
break things if pulled in now.

tim-moody:
I think of it as 'reporting' of which data-transport is an aspect.

PR #752, opened 2016-07-12
Add Reset-Network command

tim-moody:


jvonau:
Given the split should we now pull network out of roles/4-server-options/meta/main.yml?

tim-moody:
certainly worth discussing this as a strategy.  I'm not clear on the ramifications.
is there a case where a normal ansible run would pick up changes to networking without the user having to force it?

Date: Tue, 12 Jul 2016 09:07:46 -0700

jvonau:
network is currently always included in every console run, just trying to avoid picking up unintended network changes due to transient issues(cable unplugged, upstream temporary down) for something simple as disabling or enabling a service.

tim-moody:
not sure how to answer the question.  we basically never know whether network changes were intentional or not and so we don't know what to do.  so yes, maybe we should try an iteration where the user has to explicitly click a button to change the network.  it is also faster.

Date: Tue, 12 Jul 2016 09:38:26 -0700

PR #763, opened 2016-07-21
make default variable non-empty

georgejhunt:
The good news is that with this change, install-console runs to completion.


jvonau:
A bit of a brute force approach,  /admin would of been my preferred choice, maybe with "when installing", as a reminder that the gui needs to be run before any of that content is available. I use "installing" as a keyword though out the code to suppress operations during iso creation. It's just a default when the full url path is not used. I just got busy with other life obligations and didn't get a chance to do a PR after looking at Adam's machine.

PR #766, opened 2016-07-22
giving root privilege to apache is left over from my gui, and not con…

georgejhunt:
…sistent with Tim\'s console. 
Discovered this because I need to give apache rights to change iptables and conntrack info -- which might in itself be a security problem. But one step at a time. The capture portal is too big a thing to take on all at once


jvonau:
Don't think you want iptables-save, that just captures the current rules
that are already loaded. Think you would want to use iptables -I and
iptables -D to change the running rules directly. Might be best to alter a
custom named table and just use a -J(ump) to call the custom table from the
provided built-in tables.

PR #767, opened 2016-07-27
net-mods moved to 0-once need is_F18

georgejhunt:


jvonau:
George, sorry forgot about release specific code that maybe present. Please
have a look at
PR #768 which should cover all bases. Push what you feel is more
appropriate.

georgejhunt:
TK reported this. Which I fixed and pulled into release with another PR

PR #771, opened 2016-07-31
Squid captive portal

georgejhunt:


m-anish:
Pulling this into release-6.1 for testing, and running install-console, I get this error:

```
TASK: [1-prep | set_fact ] **************************************************** 
fatal: [127.0.0.1] => One or more undefined variables: 'dict object' has no attribute 'squid_dir'

```

jvonau:
whenever something in install console changes the variable gui_ver in roles/xsce-admin/defaults/main.yml needs to be bumped to re-run the already completed parts.

m-anish:
I also renamed/removed /etc/xsce/xsce.env and that seemed to do the trick

m-anish:
Okay, some thoughts on the actual PR. 

This sort of works. If I set capture_enabled = True and rerun the network setup, the following scenario works:
1. If I open a website over port 80 (HTTP), I get redirected to the schoolserver homepage.
2. If I open a website over port 443 (HTTPS), I get redirected to the schoolserver homepage. 

However, after this initial captive portal page, https doesn't work, probably because squid isn't setup to handle https

If I remove the https iptables line:  `$IPTABLES  -t nat  -A PREROUTING -i $lan -p tcp --dport 443 ! -d 172.18.96.1 -j DNAT --to 172.18.96.1:3128` then
1. The captive portal works over port 80
2. It does not work over 443

HTTPS (443) traffic works afterwards. 

So, I think before this goes in.
We should make squid transparently handle https traffic. Searching the internet, I got this:
http://rahulpahade.com/content/squid-transparent-proxy-over-ssl-https

Hope this helps.

jvonau:
something just feels icky about a man-in-the-middle squid setup for https to me.

m-anish:
Yup agree. 

Other ideas?

m-anish:
Also, anyone have experience with this?

https://msdn.microsoft.com/en-us/library/windows/hardware/dn408679.aspx

Perhaps embedding this small snippet of xml in the html header can popup a notification in phones. 

Also found this:

http://www.squid-cache.org/mail-archive/squid-users/201405/0285.html

tim-moody:
Does this interact with squid-xs.conf.j2:


{% if gw_block_https == True %}
#acl Safe_ports port 443         # https
{% else %}
acl Safe_ports port 443         # https
{% endif %}

PR #773, opened 2016-08-02
Merge2

georgejhunt:
give this a look over. I tried to be careful, but it was pretty tedious, and I may have slipped up somewhere.
Looks like this will merge with master.


georgejhunt:
I should have identified this as the PR(773) we discussed on the last call.  I includes no new code, and just tries to reconcile conflicts which develop when trying to merge release-6.1 into master.
If there was anything in master that was in conflict during the merge, it will have disappeared from master.

tim-moody:
Is this a case of reverting master to the point at which 6.1 branched off and then merging 6.1 or something more complicated?

georgejhunt:
I remembered that at one time we said some new things should go in master,
but not release-6.1. So I did the following:

   - git checkout master
   - git checkout -b merge2
   - git merge release-6.1
   - <went through all the conflicts that were thus created -- cloosing
   release-6.1>
   - git push ghunt merge2

For some of Jerry's installer, I'm not sure I chose correctly.  But I
decided to do the same thing whenever there was a conflict.

jvonau:
This seems pretty sane and correct, I'll handle the upgrade path from 6.1
to master. Should be less painful creating 6.2 or 7.0 in the future.

PR #774, opened 2016-08-23
Vnc console

georgejhunt:
I created this PR against release-6.1, but am not suggesting that it be part of 6.1.  It's just that when I create it against master, the PR becomes gibberish (because we have not yet reconciled the two.
Let's discuss at out next call -- which is guess is 4pm EDT this week


georgejhunt:
New PR which incorporates suggestions

PR #786, opened 2016-09-02
Console3

georgejhunt:
On our call this morning, I agreed to make vnc run under a non-sutoer.  But I've changed my mind. This vnc is enabled behind the xsce-admin login,  vnc only listens to localhost, and the websocket listens to a port which is protected by an iptables rule which only permits forwarding from the ip address who initiated the vnc communication. I think administering the server via a smart phone is a capabiity we should permit. This PR puts the vnc button on the control page, and pops out a screen which is separate from the admin console


tim-moody:
Please remove lines 13 to 16 from cmd-service.php, which should not have any command specific logic.
see if serverInfo has what you need.

I don't see the link in utilities70.html admin tools.

I don't understand the relationship between the adm cons user and the vnc user.  The only password that matters as far as I can tell is the one internal to vnc as I can easily bypass the whole system and go directly to vnc if I know the address.

georgejhunt:
On Fri, Sep 2, 2016 at 6:12 AM, Tim Moody <notifications@github.com> wrote:

> Please remove lines 13 to 16 from cmd-service.php, which should not have
> any command specific logic.
> see if serverInfo has what you need.
>
> I don't see the link in utilities70.html admin tools.
>
I changed utilities70 back to the way it was.

> I don't understand the relationship between the adm cons user and the vnc
> user. The only password that matters as far as I can tell is the one
> internal to vnc as I can easily bypass the whole system and go directly to
> vnc if I know the address.
>
And if vnc is only listening to localhost:5901, how is knowing the the
address going to help much? The websecket proxy at port 6080, only listens
to the ip address that opened the connection via cmd-service.php.

tim-moody:
I don't think loc.hostname is reachable over the vpn for me.

fyi function startVnc assumes success and pops up the screen.   of course an error should pop up as well if there is one.

tim-moody:
The websecket proxy at port 6080, only listens
to the ip address that opened the connection via cmd-service.php.

I don't follow.  cmd-service.php doesn't open a connection.  the front end knows the server's ip by which the client reaches it.  btw it also knows the client's ip.

PR #789, opened 2016-09-04
Arm centos

georgejhunt:
This PR is NOT meant for pulliing into our repo -- but just to document how far along CentOS is when it comes to running on armv7hl. So far I have not got console running, and I've just done 'runtags download' (have not done the download2 -- unpackaged)


georgejhunt:
console does not run, missing cracklib-python

PR #795, opened 2016-09-14
Increase assessment item download timeout

m-anish:
The current value of the assessment item download timeout consistently causes a fail on my 'fast' Indian internet. I would recommend to increase the same.


jvonau:
+1 push it

PR #799, opened 2016-09-15
The overlapping text in Chrome is due to the '.content-item ul' CSS class defined in home.css

avni:
The overlapping text in Chrome is due to the '.content-item ul' CSS class defined in home.css


avni:
FYI, not able to test this fix locally as /test/ file that exists on schoolserver.xsce.org does not exist on my local instance.

tim-moody:
added to PR 798

PR #801, opened 2016-09-22
replace stale file with symlink to correct file

jvonau:
Fixes install-console on Adam's content machine


jvonau:
This might need a revision, this uses the full path which might change if the user doesn't use /opt/schoolserver/xsce/ as the base directory.

georgejhunt:
seems like long term, we need a template module statement in the console
playbook

PR #802, opened 2016-09-22
Release 6.1

georgejhunt:
I think we should pull all of release-6.1 changes into master. But my how much Tim has done in the last month. 296 files changed 2041 additions 53 deletions.

Is this ok to do?


jvonau:
Lets sync them up, then we can test on master and port as needed to
release-6.1 or 6.2

PR #803, opened 2016-09-24
Secure

georgejhunt:
There may be additional iptables stuff needed.  This is a work in progress.


tim-moody:
Status of openvpn is not part of cmdsrv code and should be in /etc/xsce

'warn' ID for div suggests any warning could go there; does it make sense to warn someone if remore access is not on?

there could be conflicting values for openvpn_enabled in config_vars and program variables and which holds sway depends on the user action taken

remote_admin_ctl should return SUCCESS or ERROR.  Probably the callback should call get status.

not sure what this means, but at least seems backwards:

 if (remote_admin_allowed){
 +     cmd_args['activate'] = 'false';
 +     remote_admin_allowed = 'false';
 +  } else {
 +     cmd_args['activate'] = 'true';
 +     remote_admin_allowed = 'true';
 +  }

georgejhunt:
I routinely push current status at the end of a session, so that I have a restore point.  There's no need for you guys to get emails about stuff that is not finished or working.

PR #804, opened 2016-10-02
Captive portal

m-anish:
**## This is a testing pull request for preview only.** 

Instructions:
- Make sure squid and dansguardian are disabled. Otherwise, the playbook wont have any effect.
- Run the playbook

Browse to any website. You should be redirected to the captive portal page. Username and password is `xsce-admin/g0adm1n`

Future work:
1. Integrate packages into xsce-extras repo
2. Edit the default captive portal page to (a) redirect by default to school.lan (b) not require username password

Feedback welcome!


m-anish:
open issues: (notes to myself)
1. ifdef in templates/chilli/* around whether wan is available or not
2. discuss when should the playbook run. Ideally in network section, but we dont setup mysql by then.

m-anish:
**Captive portal doesnt work in offline case since the DNS cannot resolve. **

Pending **todo** is to resolve all DNS requests locally if WAN happens to be offline.

m-anish:
Yay! This is now working also when the server is not connected to the internet. The solution is to create a DNS blackhole when WAN is absent, much like what is described here:

https://doc.pfsense.org/index.php/Creating_a_DNS_Black_Hole_for_Captive_Portal_Clients#Create_the_configuration_file

**This is NOT currently a part of the PR but am documenting it here!**

m-anish:
Final list of pending issues:
1. Enable radiusd before making changes to the service file in playbook
2. Discuss and include DNS blackhole for proper offline operation
3. Properly enclose in if/endif blocks the relevant sections of the files ipup.sh.j2 config.j2

georgejhunt:
we've developed a norm for services that there are installed and enabled ansible variables which govern a service.  In this playbook, the absence or dansguardian, and squid, govern disabling dhcpd, and the enabling of captive portal.  
This dependency seems obscure, and undesirable, to me

jvonau:
Anish, can you rebase off of current master and note that DB installation is now in 0-once so maybe that would be a better place for adding freeradius. I'll work with you to resolve the network logic.

tim-moody:
finally got a chance to try this.  centos vm.  I cloned branch which is the PR

TASK: [captive_portal | Install radius packages] ******************************
changed: [127.0.0.1] => (item=freeradius,freeradius-mysql,freeradius-utils)

TASK: [captive_portal | Remove the radius database if it exists] **************
failed: [127.0.0.1] => {"failed": true}
msg: unable to connect, check login credentials (login_user, and login_password, which can be defined in ~/.my.cnf), check that mysql socket exists and mysql server is running
...ignoring

TASK: [captive_portal | Create a new database with name radius] ***************
failed: [127.0.0.1] => {"failed": true}
msg: unable to connect, check login credentials (login_user, and login_password, which can be defined in ~/.my.cnf), check that mysql socket exists and mysql server is running

tim-moody:
probably need meta like

cp roles/wordpress/meta/main.yml roles/captive_portal/meta

tim-moody:
I also think it is being disabled because mysql_enable: False, which is wrong.

tim-moody:
after patching I get to 

TASK: [captive_portal | Edit freeradius service file to ensure that it starts after mysql] ***
failed: [127.0.0.1] => {"failed": true, "rc": 257}
msg: Destination /etc/systemd/system/multi-user.target.wants/radiusd.service does not exist !

tim-moody:
starting by hand

Dec 11 15:57:03 schoolserver.lan systemd[1]: Failed to start FreeRADIUS high performance RADIUS server..
-- Subject: Unit radiusd.service has failed
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
--
-- Unit radiusd.service has failed.
--
-- The result is failed.
Dec 11 15:57:03 schoolserver.lan systemd[1]: Unit radiusd.service entered failed state.
Dec 11 15:57:03 schoolserver.lan systemd[1]: radiusd.service failed.
Dec 11 15:57:03 schoolserver.lan polkitd[528]: Unregistered Authentication Agent for unix-process:29834:1206016 (system bus name :1.175, object path /org/freedesktop/PolicyKit1/Authenticatio
lines 1275-1344/1344 (END)

tim-moody:
the include of mods-enabled does not read the conf files in the right order, so sqlcounter gets read before sql which has definitions it needs.  so we need to replace the include of the dir with explicit files:

$INCLUDE mods-enabled/always
$INCLUDE mods-enabled/attr_filter
$INCLUDE mods-enabled/cache_eap
$INCLUDE mods-enabled/chap
$INCLUDE mods-enabled/detail
$INCLUDE mods-enabled/detail.log
$INCLUDE mods-enabled/dhcp
$INCLUDE mods-enabled/digest
$INCLUDE mods-enabled/dynamic_clients
$INCLUDE mods-enabled/eap
$INCLUDE mods-enabled/echo
$INCLUDE mods-enabled/exec
$INCLUDE mods-enabled/expiration
$INCLUDE mods-enabled/expr
$INCLUDE mods-enabled/files
$INCLUDE mods-enabled/linelog
$INCLUDE mods-enabled/logintime
$INCLUDE mods-enabled/mschap
$INCLUDE mods-enabled/ntlm_auth
$INCLUDE mods-enabled/pap
$INCLUDE mods-enabled/passwd
$INCLUDE mods-enabled/preprocess
$INCLUDE mods-enabled/radutmp
$INCLUDE mods-enabled/realm
$INCLUDE mods-enabled/replicate
$INCLUDE mods-enabled/soh
$INCLUDE mods-enabled/sql
$INCLUDE mods-enabled/sqlcounter
$INCLUDE mods-enabled/sradutmp
$INCLUDE mods-enabled/unix
$INCLUDE mods-enabled/unpack
$INCLUDE mods-enabled/utf8

tim-moody:
Am I right in thinking that this only works on internal wifi?  If I have a bigger machine with external hotspot am I out of luck?

m-anish:
Interesting. I didn't encounter any issue like the one mentioned in the last comment about sql/sql-counter. 

Also to answer your previous comment, it works in both cases (i) with internal wifi (ii) with a router connected via LAN

I finally have a raspberry pi 3 so can start testing later this week. However, I am all out of NUCs for now (they are all in Spiti) and it will be atleast a couple of weeks before I get more NUCs and routers. 

Thanks for the effort in trying to get this feature upstream!

PR #806, opened 2016-10-04
Elgg

georgejhunt:
When I was testing on raspbian, I found that the elgg version 2.1 that we have in release-6.1 requires php5.5.  It shows me that my testing of elgg was really only the automated testing, and elgg does throw up a http response page.
I'd like to test this on Centos, and FC22, after which perhaps it needs to go into release-6.1.


m-anish:
TASK: [elgg | download current version from our copy] ************************* 

changed: [127.0.0.1]

TASK: [elgg | Determine if software is already expanded] ********************** 
ok: [127.0.0.1]
This fails on a xsce running on el7, in which ./runansible had been run before. The log is below:
```

TASK: [elgg | Expand it to our location] ************************************** 
skipping: [127.0.0.1]

TASK: [elgg | change ownership] *********************************************** 
failed: [127.0.0.1] => {"failed": true, "path": "/opt/elgg-1.12.12", "state": "absent"}
msg: recurse option requires state to be 'directory'

```

FWIW, the actual directory I see is `/opt/elgg` and not `/opt/elgg-1.12.12
`

m-anish:
Ignore the previous comment. I needed to delete /opt/elgg for the elgg setup to actually run again. :)

It runs now :)

jvonau:
the test for downloading should depend on the presence of install.php like previously implemented.

m-anish:
Shouldn't Jerry's comment be reported as an issue to be tracked?

georgejhunt:
Can you explain more the issue, and why it is important?

PR #812, opened 2016-10-06
ansible 2.1 does not unarchive files with blanks in filenames correct…

georgejhunt:
…ly -discovered on fc23


jvonau:
I thought ansible1.9 (available for F23) was to be used, not ansible2.1 . Not sure where F24 stands

georgejhunt:
Yes, I guess this needs to go into master, and is not really needed in release-6.1. But I do think git 2.1 is in our future.

PR #815, opened 2016-10-12
Secure

georgejhunt:
Things I have not figured out:

Get the label for handle on services page to be on one line (class=col-md-6 does not work)
Get the color of the remote button on page one to toggle
Get the WAN configButtonsEvents to do error checking as intended (does not complain 127.0.0.99999)

This requires additional polish before merging into master. Please do not merge.


georgejhunt:
I want to grey out and disable the teamviewe/openvpn checkboxes while remote_admin_allowed  is false on first page

PR #818, opened 2016-10-18
F24 rpi3

jvonau:
In testing on ansible-1.9, works on ansible-2.1, need to re-confirm F24 ansible 2.1 work without yum being installed


jvonau:
This PR runs on both ansible1.9 and ansible2.1 making F24+ less of a pain to set up by not having to 'pip install ansible1.9'. Note that ansible 2.1 run much slower than 1.9. Going to wait 24 hours for comments before pushing.

PR #830, opened 2016-10-30
ansible ownership change fails -- do in install script

georgejhunt:


georgejhunt:
This one should go into 6.1

PR #839, opened 2016-10-31
Speedup3

jvonau:
replacement for PR #821


jvonau:
replaced with RP #840

PR #841, opened 2016-11-02
Speedup3

jvonau:
if it is needed, same as #839 part of #840 but against 6.1


georgejhunt:
Code base moved away from this a lot

PR #845, opened 2016-11-08
substitute bash for sh

georgejhunt:


tim-moody:
what error did this fix, as there a likely more of these?

georgejhunt:
indeed, all of the scripts in cmd-srv use sh. I only changed the ones that
Adam requested, without checking with Tim

PR #846, opened 2016-11-10
fix bash zim

floydianslips:


tim-moody:
was done along with other scripts in another PR

PR #847, opened 2016-11-10
handle case where restarted job has no cmdargs

tim-moody:


georgejhunt:
Could you also generate a PR against raspbian-6.1 also?  Thanks

PR #863, opened 2017-01-24
unarchive is broken for zips so use unzip for elgg

tim-moody:


georgejhunt:
I think there is a missing "not" in line 65 of https://github.com/XSCE/xsce/pull/863/files#diff-b85b6e1bccccd3982050b995ee7b422b. 

So many of these!

tim-moody:
don't think so.  the logic is if postgresql_enabled turn off stock postgre and turn on  postgresql-xs

though confusing, postgresql_enabled refers to postgresql-xs

PR #865, opened 2017-01-27
suppress ansible warnings tries to use file that may not exist

tim-moody:


georgejhunt:
Reading the docs, I tried removing ansible.cfg from root, and adding
environment declarations in the main yml file.

But the warnings were not suppressed. Ideas?

PR #881, opened 2017-02-06
correct and update reference to local_content and /usb url on server

georgejhunt:


georgejhunt:
forgot to pull this in, and Adam complained, so done a second time

PR #882, opened 2017-02-07
Adams fixes

georgejhunt:


tim-moody:
the docs say /PirateShare not /Piratebox/Share, but I can't find anywhere that the actual directory is specified

georgejhunt:
never got this PR pulled in, and Adam complained that it was missing, so I fixed it on the spot

PR #884, opened 2017-02-17
Docs2

georgejhunt:
a big chunk of changes, check it out first, and merge the PR into a testing context


georgejhunt:
I'm not sure this is good enough.  I've managed to get the wiki, faq, and release notes into html, but the format is not yet very good. I spent quite a bit of time trying to rewrite the links, so that they work. But I'm running out of patience, and it seems like there are higher priorities.

Tim, note that I put a link to the iiab-info on the utilities page.  Is that ok with youf?

PR #891, opened 2017-03-14
let xsce-admin be a sudoer without password

georgejhunt:
This PR creates sudo group on CentOS, for cross distribution compatibility.


tim-moody:
I am OK with this as long as it is conditional on exactly the when clause in 

roles/3-base-server/tasks/main.yml

PR #900, opened 2017-04-08
networking on nuc is broken, backport fixes from iiab repo

georgejhunt:
I think it is too late in the game to add this to 6.2. But I think we can ask TK to test this, and I find that I need this change to get my NUC image to work.  I really focused most of my attention on RPI for this release.


holta:
Thanks @georgejhunt & All for catching this!  I agree with your RPi focus.

But: am also in favor of including even risky fixes (to broaden our testing coverage!) if this stands to uphold our legacy/base community of NUC, BRIX, MSI, Zotac, etc ?

roles/1-prep/tasks/detected_network.yml
```diff
@@ -250,7 +250,7 @@
 - name: 2 or more devices on the LAN - use bridging
   set_fact:
      xsce_lan_iface: br0
-  when: 'discovered_lan_iface != "none" and num_lan_interfaces >= "2"'
+  when: 'discovered_lan_iface != "none" and num_lan_interfaces >= "1"'
```
jvonau: I'd rather see a new stanza with "when is_debian" rather than mess with the CentOs' code path where is could break working code.

holta:
Update from today's call: @georgejhunt will work on suggestion from Jerry (stanza for CentOS)

Issue #905, opened 2017-04-13
Owncloud Configure Error


floydianslips:
After running "Install Configured Options" with the following services:

Moodle
Kiwix
KAlite
KAlite Downloader
OSM
Sugarizer
OwnCloud
Munin
VnStats
AWStats

Attempting to access OwnCloud Home Page gave this error:

Error: Error while trying to create admin uswer: An exception occurred while executing 'CREATE TABLE `appconfig`(`appdid`VARCHAR(32)DEFAULT"NOT NULL, `configkey` VARCHAR(64) DEFAULT" NOT NULL, `configvalue` LONGTEXT DEFAULT NULL, INDEX appconfig_config_key_index (`configkey`), INDEX appconfig_appid_key (`appid`), PRIMARY KEY(`appid`,`configkey`)) DEFAULT CHARACTER SET utf8 COLLATE utf8_bin ENGINE = InnoDB':SQLSTATE[42S01]: Base table or view already exists: 1050 Table `appconfig` already exists

floydianslips:
Apache error is: AH01753: access check of 'localhost' to /owncloud/index.php/heartbeat failed, reason: unable to get the remote host name

Issue #904, opened 2017-04-13
generic sqlite support


tim-moody:
we should add the sqlite3 client package and php5-sqlite, maybe in common

jvonau:
Should be fixed by #906 but is not as yum.yml is not called, should be in packages.yml. confirm package name of sqlite3 across all distros please. Don't want a repeat of #918. Problem exists in [iiab](https://github.com/iiab/iiab/search?utf8=%E2%9C%93&q=sqlite3&type=) also.

PR #906, opened 2017-04-13
add in php5-curl,php5-sqlite,sqlite3,change group on workpress file

georgejhunt:


holta:
Looks like this builds upon #903 #904 

And possibly also #905 (or @floydianslips can clarify if that's entirely different?)

Am testing it on Pixel on RPi3 now, Thx @georgejhunt

holta:
Clarif (building off TK's WordPress work in #903) just for the record:

Line 17 was fixed from:
chown root:apache $DEST

...over to:
chown root:{{ apache_data }} $DEST

Inside this file:
https://github.com/XSCE/xsce/blob/release-6.2/roles/wordpress/templates/get-xsce-wp-salts.j2

Based on Ansible variable {{ apache_data }} which is set to www-data here:
https://github.com/XSCE/xsce/blob/release-6.2/vars/Debian.yml

(Gracias @tim-moody  for catching this during this morning's call @ http://tinyurl.com/iiabminutes)

PR #907, opened 2017-04-13
convert job output to html safe

tim-moody:


tim-moody:
tested on nuc debian

tim-moody:
doesn't fix it

PR #909, opened 2017-04-18
test whether removing docs from index will fix git pull

georgejhunt:
I was able to do a xsce pull, onto this test branch, and get an "up to date" response an no conflicts


tim-moody:
why not just have refresh-wiki-docs.sh copy to /library/www/html/info

PR #910, opened 2017-04-19
alternative to refresh-wiki-docs.sh that copies to /library/www/html/info

tim-moody:


tim-moody:
there are some links that are not rewritten
did not remove the apache conf file for iiab-info

PR #913, opened 2017-05-13
move br0 to interfaces, disable resolvconf

georgejhunt:
testing to date:
applied to Josh's pixel and pixel-lite images. results:
   pixel: reboot went from 3:02 to 0:44
   lite: reboot went from 3:55 to 0:31
without ethernet: pixel reboots in 0:31
"box" and "box.lan" still resolve on attached client

No testing on other-than-rpi

### DIFF
--- a/roles/1-prep/tasks/computed_vars.yml
+++ b/roles/1-prep/tasks/computed_vars.yml
@@ -38,8 +38,6 @@
       value: '{{ ansible_local["local_facts"]["xsce_branch"] }}'
     - option: 'runtime_commit'
       value: '{{ ansible_local["local_facts"]["xsce_commit"] }}'
-    - option: 'xsce_uuid'
-      value: '{{ ansible_local["local_facts"]["xsce_uuid"] }}'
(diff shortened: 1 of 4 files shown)


tim-moody:
for gateway you changed eth0 from dhcp to manual. are you sure you get an ip address when it is connected?

georgejhunt:
That seems to be the whole function of dhcpcd. But specifying manual, the
networking.service is instructed to let dhcpcd do it.

jvonau:
Please remove the now unused template files. Should the APPLIANCE section receive the same treatment? Sounds more like a race for control of the network card. Something is not quite right, network.service using dhcp in the interface file should call a dhcp client daemon without manually enabling the dhcpdcd service. If the dhcpcd.service is enabled and has been started before network.service then there will be a stall while network.service is trying to enable and the take control of the card then can't. Same holds true during shutdown. Have a closer look at the logs.

jvonau:
There is a typo in rpi_debian.yml also:  "command: ifdown etho"

PR #918, opened 2017-06-18
centos fix

georgejhunt:
I discovered that the centos install works with more recent pip and setup tools -- so I modified http://download.iiab.io/6.2/x86/centos-load.txt to include "pip install -U pip setuptools"

### DIFF
--- a/roles/httpd/tasks/main.yml
+++ b/roles/httpd/tasks/main.yml
@@ -17,7 +17,8 @@
     - httpd
     - php
     - php-curl
-    - php-sqlite
+# centos does not have
+#    - php-sqlite
   tags:
     - download
   when: not is_debian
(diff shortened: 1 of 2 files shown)


jvonau:
Interesting that one needs to update pip after installing pip from scripts/ansible. Just feels like the wrong place to run the update from. Just my 2 cents.

georgejhunt:
Following the suggestion at
https://packaging.python.org/tutorials/installing-packages/#id9

jvonau:
Sorry I just figured the most recent "easy_install and pip" would be installed by scripts/ansible for both CentOS and Debain from the start, without the need to do so from an external script.

tim-moody:
welcome the day when the latest ansible works for us and this kluge is not
needed.

jvonau:
Think the "is_*distro* booleans" in vars/default_vars|*distros*.yml might need the same treatment as in https://github.com/iiab/iiab/commit/545b04829471b79e96652ef8abff5ba39e190cc0

PR #921, opened 2017-06-26
reviewing

jvonau:
don't merge


jvonau:
do-able but not worth my time

